PR #6, opened 2025-06-03
feat: use file as locks for snap management

IronCore864:
Edit from @lucabello: I know it says don't merge, and this is now merged. Scroll to the [bottom](https://github.com/canonical/opentelemetry-collector-operator/pull/6#issuecomment-3000773802)!

---

Proof of concept, _do not merge_.

This PR uses files as a lock and reference counter when handling snaps. Each unit "registers" itself with a snap when installing and "unregisters" itself from that snap when uninstalling. The snap would only be removed if there are no other units registered with the snap.

See the [other PR](https://github.com/canonical/opentelemetry-collector-operator/pull/5) that uses SQLite for locking.

As discussed elsewhere, we prefer using file locks or flock when merging configurations (not done yet), so that we don't have to introduce a database.


src/snap_lock.py
```diff
@@ -0,0 +1,119 @@
...
+        """Returns True if the snap is used by other instances besides the current instance, False otherwise."""
+        for filename in os.listdir(self.lock_dir):
+            if filename.startswith(snap_name + self.separator):
+                instance_name = filename[len(snap_name + self.separator) :]
+                if instance_name != self.instance_name:
+                    return True
+        return False
+
+    @contextmanager
+    def lock_snap(self, snap_name: str, timeout: int = 3) -> Generator[None, None, None]:
```
sed-i, line 83: Snap removal may take longer than 3 seconds. Should we increase to 30 just to be safe?

src/snap_lock.py
```diff
@@ -0,0 +1,119 @@
...
+class SnapLock:
+    """SnapLock provides mechanisms for singleton snaps using fcntl.flock and file-based reference counting.
+
+    Raises SnapLockError on any error.
+    """
+
+    def __init__(self, instance_name: str, lock_timeout_sec: int = 300):
+        # Replace all non-alphanumeric chars (except _-) with _
+        self.instance_name = re.sub(r'[^\w-]', '_', instance_name)
+        self.lock_timeout = lock_timeout_sec
```
sed-i, line 26: Seems unused.

src/snap_lock.py
```diff
@@ -0,0 +1,119 @@
...
+        self.separator = '--'
+
+        self.lock_dir = os.path.join('/run/lock', 'snap_locks')
```
sed-i, line 29: I came across [this](https://systemd.io/TEMPORARY_DIRECTORIES/) the other day, but not sure how to correctly implement 

> use your own sub-directory under /run/

Your impl seems right, just wanted to throw the ref, but you probably already know of it :)

src/snap_lock.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            os.makedirs(self.lock_dir, exist_ok=True)
+            os.chown(self.lock_dir, os.geteuid(), os.getegid())
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise SnapLockError(f'Error creating lock directory: {e}') from e
+
+    def _get_lock_file_path(self, resource_name: str) -> str:
+        return os.path.join(self.lock_dir, f'{resource_name}.lock')
```
sed-i, line 38: Could we narrow down the terminology? Currently we have
- lock dir and lock file
- resource name
- snap name
- reference file
- instance name

I am confused :)

sed-i (commented):
Looks very nice!

src/charm.py
```diff
@@ -35,16 +43,38 @@ def _install(self) -> None:
...
+                )
 
     def _remove(self) -> None:
         if self.hook != "remove":
             return
 
+        lock = SnapLock(self.unit.name)
         for snap_package in SNAPS:
-            self._remove_snap(snap_package)
+            lock.unregister(snap_package)
```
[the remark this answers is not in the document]
IronCore864: I think not, because they are two separate actions. Unregister decreases the counter (in this case, it removes a file); and when removing a snap, it acquires a different lock (a different file) to make sure there is only one removing in process.

src/snap_lock.py
IronCore864: According to this article from systemd, it seems a subdirectory under `/run/` is preferred. According to Linux conventions, it seems we should use `/var/lock` (linked to `/run/lock`) for lock files. Combining these two, we end up with `/run/lock/some_subdirectory` is preferred. Also, I noticed that lock files have a naming convention that starts with "LCK..". I will also make this change.

MichaelThamm:
From a conversation with the team, we need to think through the following 2 scenarios more:
- Charm upgrade/downgrade
- Another charm (which installs the node-exporter and/or otelcol snap) installed on the same machine will overwrite our snaps

tests/unit/test_singleton_snap.py
```diff
@@ -0,0 +1,138 @@
...
+    else:
+        pytest.fail('Lock file was not created in time')
+
+    p2.start()
+    p1.join()
+    p2.join()
+
+    results = [acquired.get(), acquired.get()]
+    assert results.count(True) == 1
+    assert results.count(False) == 1
```
sed-i, line 138: Should we scale up the concurrent test to e.g. 20 processes?

README.md
```diff
@@ -1 +1,49 @@
-# opentelemetry-collector-operator
\ No newline at end of file
+# opentelemetry-collector-operator
+
+## SingletonSnapManager: Locking and Registration
```
sed-i, line 3: This is a great piece of documentation!
Since this is a charm repo, should we move this information into the `src/singleton_snap.py` module level docstring?

src/charm.py
```diff
@@ -10,63 +10,90 @@
...
+    install_snap,
+    node_exporter_snap_name,
+    opentelemetry_collector_snap_name,
+)
+from singleton_snap import SingletonSnapManager
 
 # Log messages can be retrieved using juju debug-log
 logger = logging.getLogger(__name__)
-VALID_LOG_LEVELS = ["info", "debug", "warning", "error", "critical"]
+VALID_LOG_LEVELS = ['info', 'debug', 'warning', 'error', 'critical']
```
sed-i, line 25: @lucabello mind checking why our `tox -e fmt` is doing this here?

src/charm.py
```diff
@@ -10,63 +10,90 @@
...
-        if self.hook != "remove":
+        if self.hook != 'remove':
             return
 
+        manager = SingletonSnapManager(self.unit.name)
         for snap_package in SNAPS:
-            self._remove_snap(snap_package)
+            manager.unregister(snap_package)
+            if not manager.is_used_by_other_units(snap_package):
+                with manager.snap_operation(snap_package):
```
sed-i, line 73: From first glance seems like we need to swap these two lines to avoid race condition.

sed-i (commented):
Great job @IronCore864 !

tests/unit/test_singleton_snap.py
```diff
@@ -0,0 +1,138 @@
...
+from multiprocessing import Process, Queue
+from src.singleton_snap import SingletonSnapManager, SingletonSnapError
+
+
+@pytest.fixture
+def lock_dir(monkeypatch):
+    tmp = tempfile.mkdtemp()
+    monkeypatch.setattr(SingletonSnapManager, 'LOCK_DIR', tmp)
+    yield tmp
+    shutil.rmtree(tmp)
```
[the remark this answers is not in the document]
IronCore864: After thinking about this, I will change it to pytest's tmp_dir fixture, something like:

```python
@pytest.fixture
def lock_dir(monkeypatch, tmp_path):
    lock_dir = tmp_path / "lock_dir"
    lock_dir.mkdir()
    monkeypatch.setattr(SingletonSnapManager, 'LOCK_DIR', lock_dir)
    yield lock_dir
```

Because I believe this is the most robust and pytest-y solution, the directory will be automatically cleaned up after the test session ends, regardless of the test's success or failure.

tests/unit/test_singleton_snap.py
```diff
@@ -0,0 +1,138 @@
...
+        time.sleep(sleep_time)
+
+
+def test_lock_timeout(lock_dir):
+    # Start a process that holds the lock.
+    p = Process(target=hold_lock, args=('unit-1', 2, lock_dir))
+    p.start()
+
+    # Wait until the lock file appears to ensure the lock is held, up to 1 second.
+    lock_file = f'{lock_dir}/LCK..otelcol'
```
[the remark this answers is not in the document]
IronCore864: This is because `SingletonSnapManager.LOCK_FILE_PREFIX` is set to `LCK..`. I noticed that the naming convention dictates that the lock file name must be "LCK.." followed by something else, see [here](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s09.html#:~:text=The%20naming%20convention%20which%20must,HDB%20UUCP%20lock%20file%20format.).

tests/unit/test_singleton_snap.py
IronCore864: Good idea, will do.

src/charm.py
```diff
@@ -10,63 +10,90 @@
...
+                # This is a placeholder for actual configuration merging logic.
+                # For example:
+                #
+                # content = merge_config()
+                # with open('etc/otelcol/config.yaml', 'w') as f:
+                #     f.write(content)
+                #     f.flush()
+                pass
 
     def _remove(self) -> None:
```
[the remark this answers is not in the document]
IronCore864: I did some local testing and it seems stop/remove both are OK in this usecase, but according to [Juju doc](https://documentation.ubuntu.com/juju/3.6/reference/hook/#remove), the stop event is emitted prior to remove, and all necessary steps for handling removal should be handled there. So I'm changing this from remove to stop.

src/singleton_snap.py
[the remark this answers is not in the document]
IronCore864: If it's required in other places, I think we can add a `namespace` parameter with a default value of "singleton_snaps" in the `__init__`, maybe even a `lock_dir_base` that defaults to `/run/lock/`, then we can do `self.LOCK_DIR = os.path.join(lock_dir_base, self.namespace)`. In this case, the class should not be named `SingletonSnapManager` anymore. For now, I will not future-proof it in this PR.

src/singleton_snap.py
```diff
@@ -0,0 +1,203 @@
...
+                    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                    break
+                except BlockingIOError:
+                    if time.time() > deadline:
+                        raise SingletonSnapError(f'Timeout acquiring lock for {lock_name}')
+                    time.sleep(0.1)
+            try:
+                yield
+            finally:
+                fcntl.flock(f, fcntl.LOCK_UN)
```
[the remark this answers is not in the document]
IronCore864: 1. Should LOCK_UN be closer to LOCK_EX?

I think it might not be necessary. Here I acquire a lock (LOCK_EX) before the critical section that is yield, and release it (LOCK_UN) after the section is done, and it's in the finally block, so it's guaranteed.

If we want to move it closer, I have another draft that is identical:

```python
        lock_path = os.path.join(self.LOCK_DIR, f'{self.LOCK_FILE_PREFIX}{lock_name}')
        with open(lock_path, 'w') as f:
            deadline = time.time() + timeout
            while True:
                try:
                    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)

                    try:
                        yield
                    finally:
                        fcntl.flock(f, fcntl.LOCK_UN)
                        try:
                            os.remove(lock_path)
                        except FileNotFoundError:
                            pass
                        except OSError:
                            raise SingletonSnapError(f'Error removing lock file for {lock_name}')

                    break
                except BlockingIOError:
                    if time.time() > deadline:
                        raise SingletonSnapError(f'Timeout acquiring lock for {lock_name}')
                    time.sleep(0.1)
```

But I'm not sure if this more readable.

2. Should `fcntl.flock(f, fcntl.LOCK_UN)` be inside the try?

I think it's best to place LOCK_UN in a finally block. This ensures the lock is always released, even if an exception occurs.

If you mean the `flock(..., LOCK_EX)`, I think it's in the right place, the outer try handles race conditions/timeout, and the inner try makes sure it's always unlocked.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+        self._update_registration(snap_name, create=False)
+
+    @contextmanager
+    def _lock_directory(self):
+        dir_fd = None
+        try:
+            dir_fd = os.open(self.LOCK_DIR, os.O_RDONLY)
+            fcntl.flock(dir_fd, fcntl.LOCK_SH)
+            yield
+        except TypeError as e:
```
benhoyt, line 175: It seems odd to me to be catching a `TypeError` (programmer error) and re-raising it as a `SingletonSnapError`. Is it better to just let type errors bubble up?

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+            raise SingletonSnapError(f"Invalid lock directory path: {e}") from e
+        except FileNotFoundError as e:
+            raise SingletonSnapError(f"Lock directory not found: {e}") from e
+        except OSError as e:
+            raise SingletonSnapError(f"Error locking directory: {e}") from e
+        finally:
+            if dir_fd is not None:
```
benhoyt, line 182: Can this ever happen? It seems to me `dir_fd` will be set, or an exception will be raised. We can probably drop this and the `dir_fd = None` line at the start.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+                try:
+                    fcntl.flock(dir_fd, fcntl.LOCK_UN)
+                    os.close(dir_fd)
+                except Exception as e:
```
benhoyt, line 186: Better to restrict it to `OSError` here? Otherwise we'll catch accidental spelling mistakes and so on.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Raises:
+            SingletonSnapError: If an OS error occurs.
+        """
+        lock_path = self._get_registration_file_path(snap_name)
+        action = "registering" if create else "unregistering"
+
+        try:
+            with self._lock_directory():
+                if create:
```
benhoyt, line 203: This function has two different modes and if/else between them (another if/else in the `action = ...` line above). Would it be better to just include the correct code directly in register/unregister?

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+                    open(lock_path, "w").close()
+                else:
+                    try:
+                        os.remove(lock_path)
+                    except FileNotFoundError as e:
+                        # pass
```
benhoyt, line 209: Should this be commented out? Or should the raise be `pass` instead?

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+                        raise SingletonSnapError(f"Error {action} unit: {e}") from e
+        except Exception as e:
```
benhoyt, line 211: This is too broad an exception class to catch. Can we restrict at least to `OSError`? Otherwise it'll catch all manner of things like `NameError` and `AttributeError`.

That makes me think, though. Do users of this module really care about catching `SingletonSnapError`? Maybe we should just let all errors bubble up instead?

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        prefix = f"{self.LOCK_FILE_PREFIX}{self._normalize_name(snap_name)}{self.SEPARATOR}"
+        units = []
+
+        try:
+            with self._lock_directory():
+                for filename in os.listdir(self.LOCK_DIR):
+                    if filename.startswith(prefix):
+                        units.append(filename[len(prefix) :])
+        except Exception as e:
```
benhoyt, line 234: Too broad again -- let's use `OSError` or `SingletonSnapError` or similar.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+                else:
+                    try:
+                        os.remove(lock_path)
+                    except FileNotFoundError as e:
+                        # pass
+                        raise SingletonSnapError(f"Error {action} unit: {e}") from e
+        except Exception as e:
+            raise SingletonSnapError(f"Error {action} unit: {e}") from e
+
+    def get_units(self, snap_name: str) -> List[str]:
```
benhoyt, line 214: It seems to me this function shouldn't exist, as it's always going to be racy. You might call it, and another unit might register at the same time, so your result is wrong.

Alternatively, we should not acquire the lock inside here, and document that it must be called with the lock held (for a group of atomic operations).

Similar for `is_used_by_other_units` -- should only be called with the lock held for a group of operations, otherwise its value is racy.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+            raise SingletonSnapError(f"Error reading unit list: {e}") from e
+
+        return units
+
+    def is_used_by_other_units(self, snap_name: str) -> bool:
+        """Check if snap is being used by other units."""
+        return any(unit != self.unit_name for unit in self.get_units(snap_name))
+
+    @contextmanager
+    def _acquire_lock(self, lock_name: str, timeout: int) -> Generator[None, None, None]:
```
benhoyt, line 244: Normally timeouts are done as float seconds in Python (which can be an int as well), so you can do 0.5 or similar.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+        """Internal method to acquire an exclusive lock with timeout.
+
+        Uses fcntl.flock to acquire an exclusive lock on a lock file.
+        Retries until the timeout is reached.
+        """
+        lock_path = os.path.join(self.LOCK_DIR, f"{self.LOCK_FILE_PREFIX}{lock_name}")
+        with open(lock_path, "w") as f:
+            deadline = time.time() + timeout
```
benhoyt, line 252: Better to use [time.monotonic](https://docs.python.org/3/library/time.html#time.monotonic) for this sort of thing. Same below.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+            while True:
+                try:
+                    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                    break
+                except BlockingIOError:
+                    if time.time() > deadline:
+                        raise SingletonSnapError(f"Timeout acquiring lock for {lock_name}")
```
benhoyt, line 259: If we get rid of `SingletonSnapError` (see above), this could just be `TimeoutError(<message>)`.

src/singleton_snap.py
```diff
@@ -0,0 +1,293 @@
...
+            try:
+                yield
+            finally:
+                fcntl.flock(f, fcntl.LOCK_UN)
+                try:
+                    os.remove(lock_path)
+                except FileNotFoundError:
+                    pass
+                except OSError:
+                    raise SingletonSnapError(f"Error removing lock file for {lock_name}")
```
benhoyt, line 270: Here we're throwing away the original exception. Should at least do `from e` (though per above, I'd suggest just letting the original `OSError` bubble up, not catching it at all.

benhoyt (commented):
Nice, this looks pretty close! I think a bit more thought should be given to exception handling. I'd recommend getting rid of the custom exception type altogether, and just letting the original errors bubble up (if the caller isn't going to catch them anyway). And in some cases you're losing information / catching too much.

src/singleton_snap.py
IronCore864: It won't happen; it is only to suppress static check warnings. I will remove them and add two lines of `# type: ignore`.

src/singleton_snap.py
IronCore864: I had included that in the first draft, then refactored it into this version to simplify it, but I agree that it now has too many levels of indentation, which might not be an improvement in terms of readability. I will change it back.

src/singleton_snap.py
IronCore864: Thanks for pointing it out. Originally, I thought we could just pass, because if the file is already removed, it was kind of OK. Leon pointed out that that might cover some implementational error, so it's better not to pass.

src/singleton_snap.py
benhoyt: Yeah, that's fair -- might be good to document that it's primarily for debugging.

src/charm.py
```diff
@@ -10,41 +10,68 @@
...
             return
 
+        manager = SingletonSnapManager(self.unit.name)
         for snap_package in SNAPS:
-            self._install_snap(snap_package)
-            self._start_snap(snap_package)
-
-    def _remove(self) -> None:
-        if self.hook != "remove":
+            manager.register(snap_package)
```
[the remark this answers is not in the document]
IronCore864: I see the value in it, and I will add it. It's fairly simple since we already have a lock "file": we can put the revision in the file. Then `get_revisions` can just parse all revisions from the files.

Note that currently the revision is hard-coded in `snap_maps` in `snap_management.py`,  I did not change this.

IronCore864:
Most comments resolved, doc updated as discussed, ready to be reviewed for a last time before merging.

src/charm.py
```diff
@@ -10,41 +10,68 @@
...
+            with manager.snap_operation(snap_package):
+                self._install_snap(snap_package)
```
sed-i, line 50: We will need to use `get_revisions()` somewhere, and include logic such as:

```
if our_revision > max(get_revisions()):
    self._install_snap()
```

src/charm.py
```diff
@@ -10,41 +10,68 @@
...
+                self._start_snap(snap_package)
+
+            with manager.config_operation(snap_package):
+                # Merge configurations under a directory into one,
+                # and write it to the default otelcol config file.
+                # This is a placeholder for actual configuration merging logic.
+                # For example:
+                #
+                # content = merge_config()
```
sed-i, line 59: Silly thought: Would it be useful if the `registration_file_path` we use per unit actually contains the config itself?
Because we need some prescribed location for the individual config anyway, and since we already have such a prescribed path, then perhaps we could make use of it for the per-unit config?

IronCore864: I think it is a good idea, we need a place to store the individual config anyway

src/charm.py
lucabello: Keeping it the way we have it in all charms for now

src/singleton_snap.py
```diff
@@ -0,0 +1,203 @@
...
+    @staticmethod
+    def _normalize_name(name: str) -> str:
+        """Normalize names to contain only alphanumerics, _ and -."""
+        return re.sub(r'[^\w-]', '_', name)
+
+    def _ensure_lock_dir_exists(self) -> None:
+        """Ensure the lock directory exists with correct permissions."""
+        try:
+            os.makedirs(self.LOCK_DIR, exist_ok=True)
+            os.chown(self.LOCK_DIR, os.geteuid(), os.getegid())
```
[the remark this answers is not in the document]
lucabello: @sed-i

src/charm.py
lucabello: Done here: [`87c9988` (#6)](https://github.com/canonical/opentelemetry-collector-operator/pull/6/commits/87c99886dd3b1da510a4cba4919b62c8a2751dd7)

src/charm.py
lucabello: @sed-i the `registration_file_path` contains the snap revision:

```python
    def register(self, snap_name: str, revision: int) -> None:
        """Register current unit as using the specified snap and revision.

        Args:
            snap_name: Name of the snap.
            revision: Optional revision to put in the lock file. Defaults to an empty string.

        Raises:
            OSError: if there is an I/O related error creating the lock file.
        """
        lock_path = self._get_registration_file_path(snap_name)
        with self._lock_directory():
            with open(lock_path, "w") as f:
                f.write(str(revision))
```

Where do we put the config?

lucabello:
CI is passing, but CLA/Check libraries isn't :( Merging!

PR #13, opened 2025-06-26
feat: remove locking mechanism and refactor unclear patterns

### LINKED ISSUE #12: Remove locking mechanism from the charm
### Enhancement Proposal

We should keep the snap registration but remove the locking mechanism, as we found out it's not necessary because Juju guarantees hooks are executed one by one.

lucabello:
## Issue
Closes #12 

## Solution
This PR not only removes the locking mechanism, but also simplifies the snap management process and makes it more robust. 

In preparation of the config management logic, the snap revision was added to the filename; in a future PR we'll probably not write it inside the file as well (as that might hold the workload config).

I kept the `hook` matching logic and just made it more visible, even though I'm not sure whether we should keep it or not (in any case, it wouldn't be changed in this PR).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
-# Copyright 2025 jose
+# Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
-#
-# Learn more at: https://juju.is/docs/sdk
 """A Juju charm for OpenTelemetry Collector on machines."""
 
 import logging
@@ -14,11 +12,8 @@ from snap_management import (
     SnapSpecError,
     SnapInstallError,
     SnapServiceError,
-    get_system_arch,
     install_snap,
-    node_exporter_snap_name,
-    opentelemetry_collector_snap_name,
-    snap_maps,
+    SnapMap,
 )
 from singleton_snap import SingletonSnapManager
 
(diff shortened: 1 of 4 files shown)


MichaelThamm (approved):
Reviewed in our pair programming. LGTM

src/charm.py
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
-# Copyright 2025 jose
+# Copyright 2025 Canonical Ltd.
```
dimaqq, line 2: 👏🏻

dimaqq (commented):
I like the cleanup!

There's one potential issue that keeps nagging me: what is the unit is forcibly removed or unit's code has a major bug or some kind of situation where charm's remove event is not processed or fails.

Then that unit's file remains in the directory forever.

I've solved a similar problem recently by writing some process id in the the file, and occasionally checking if that pid is still alive.
https://github.com/dimaqq/tracing-integration-tester/blob/dc6a3b410b22c25534a42e7ab5b07d577a5be870/src/server.py#L138-L143

I suppose in the charm's case, it would have to be the unit agent process id, not the charm's.

Alternatively, each unit could snoop other unit names on the peer relation, and if it sees a file without the matching peer unit, it would just go ahead and remove it? Maybe with a good delay to work around data races.

What do you reckon?

PR #14, opened 2025-06-30
feat: add otelcol workload and cos-agent integration

lucabello:
## Issue
The current charm is a skeleton that only handles snap installs and removals. We need to add an actual workload implementation for otel collector, plus support for the COS Agent library.

Supersedes #8 

## Solution
This PR adds the actual otel collector workload, and support for COS Agent (only metrics for now; logs can be uncommented once we have the correct receiver in the snap).

Note that the COS Agent integrations isn't polished yet; I'd like to complete it (including logs) before refactoring it to a separate function.

## Context
The code was moved from the k8s charm. The only differences are contained in `charm.py`.

Technically, `constants.py` also has changes due to the certs needing to be saved under `/var/snap/opentelemetry-collector/common` to be readable by the snap.

## Testing Instructions
Currently, I have only manually tested this charm. We could write tests in a separate PR so that it's more self-contained.

Screenshot-2025-06-30_14-30

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -229,6 +229,8 @@ class OpentelemetryCollectorOperatorCharm(ops.CharmBase):
         tracing_otlp_http_endpoint = integrations.send_traces(self)
         if tracing_otlp_http_endpoint:
             config_manager.add_traces_forwarding(tracing_otlp_http_endpoint)
+        ## COS Agent tracing
+        cos_agent.update_tracing_receivers()
 
         # Dashboards setup
         ## COS Agent dashboards


lucabello:
Correct, the only file to look at is `charm.py` :)

sed-i:
> charm status is set to maintenance to "install" both snaps

Not sure what "both snaps" mean, but changing status for a seconds is not a big deal?

> I think we can do that in a follow-up PR maybe?

Sure.

lucabello:
I fixed some things like alerts and dashboards not being sent.

Now that the `filelog` receiver is in the snap, I'm trying to get logs to work as well. It seems like the FStab file is created and populated, but it's not mounted correctly. Do we do the mounting manually in grafana-agent? Or is the FStab mounted automatically? Either way, the logs are not showing up in the expected folder.

lucabello:
Logs are now working:
Screenshot-2025-07-03_12-43

PR #23, opened 2025-07-15
feat: handle multiple configs for colocated principals

### LINKED ISSUE #19: Merge configs from multiple principals
### Enhancement Proposal

When multiple otel collectors are on the same machine (which happens when multiple principals are on the same machine), the configs need to be merged.

Currently, whichever unit executes an event last overrides the `/etc/otelcol/config.yaml` file. 

We should use the "registration file" (the one made by `singleton_snap.py`) to store the config generated by each otelcol unit, then have a function to merge them.

lucabello:
## Issue
Closes #19 

## Solution
Detailed in [this comment](https://github.com/canonical/opentelemetry-collector-operator/issues/19#issuecomment-3077772565).

## Context
Supporting multiple principals on the same machine.

## Testing Instructions
Unit tests should be added after the logic is decided on.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -302,7 +302,8 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
 
         # TODO: Conditionally open ports based on the otelcol config file rather than opening all ports
         # Append port 9100 for Node Exporter # TODO: is this needed?
-        self.unit.set_ports(*[port.value for port in Port])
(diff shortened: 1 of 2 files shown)


lucabello:
This currently adds the "writing config to the registration file" functionality. We should probably remove the `SingletonConfigManager` class I added and follow the approach I detailed in the issue instead.

Issue #11, opened 2025-06-20
Charm's status should reflect mandatory relation pairs requirements


sed-i:
### Enhancement Proposal

(VM side of https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/67)

Charm's status should inform about missing manadatory relation pairs.

Ref:
- https://github.com/canonical/grafana-agent-operator/pull/85
- https://github.com/canonical/cos-lib/blob/6cd3769727419f50018994591a90c0024542389b/src/cosl/mandatory_relation_pairs.py#L9

lucabello:
We should, instead of using relation names to check for mandatory relation pairs, check whether the only exporter set for a pipeline is the debug exporter - if so, set `BlockedStatus()`.

PR #34, opened 2025-07-22
feat: Status message on missing relation pairs

MichaelThamm:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/11

## Solution

1. BlockedStatus if missing sink relations when a source relation exists.
2. We do not care if a sink relation exists without a source relation.

## Testing Instructions

`tox -e unit -- tests/unit/test_mandatory_relation_pairs.py`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -371,6 +371,11 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
 
         self.unit.status = ActiveStatus()
 
+        # Mandatory relation pairs
+        missing_relations = integrations.get_missing_mandatory_relations(self)
+        if missing_relations:
+            self.unit.status = BlockedStatus(missing_relations)
+
     def _install(self) -> None:
         manager = SingletonSnapManager(self.unit.name)
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -47,13 +47,19 @@ async def charm(ops_test: OpsTest) -> str:
     our CI will currently set the variable to the highest-base one.
     """
     if charm_file := os.environ.get("CHARM_PATH"):
-        # Make sure we're using '22.04' in tests, because that's Zookeeper's base
-        charm_file = charm_file.replace("24.04", "22.04")
-        return str(charm_file)
+        return charm_file
 
     charm = await ops_test.build_charm(".")
     assert charm
     return str(charm)
+
+
+@pytest.fixture(scope="module")
+@timed_memoizer
+async def charm_22_04(charm) -> str:
+    """Charm (platform = ubuntu@22.04) used for integration testing."""
+    # Note: Use '22.04' in integration tests with Zookeeper, because that's Zookeeper's base
+    return charm.replace("24.04", "22.04")
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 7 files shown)


src/charm.py
```diff
@@ -371,6 +371,11 @@ def _reconcile(self):
 
         self.unit.status = ActiveStatus()
 
+        # Mandatory relation pairs
+        missing_relations = integrations.get_missing_mandatory_relations(self)
+        if missing_relations:
+            self.unit.status = BlockedStatus(missing_relations)
```
Abuelodelanada, line 377: Small suggestion to reduce lines of code :P

```suggestion
        if missing_relations := integrations.get_missing_mandatory_relations(self):
            self.unit.status = BlockedStatus(missing_relations)
```

Abuelodelanada (approved):
LGTM Mike!

lucabello:
Some notes:
- for the mandatory relation pairs we need, just mimic grafana-agent
- the snap management fix is needed for itests to pass; it's a leftover, we don't read revisions from that "lockfile" contents anymore, but from its name

tests/integration/conftest.py
```diff
@@ -47,15 +47,21 @@ async def charm(ops_test: OpsTest) -> str:
...
     """
     if charm_file := os.environ.get("CHARM_PATH"):
-        # Make sure we're using '22.04' in tests, because that's Zookeeper's base
-        charm_file = charm_file.replace("24.04", "22.04")
-        return str(charm_file)
+        return charm_file
 
     charm = await ops_test.build_charm(".")
     assert charm
     return str(charm)
```
MichaelThamm, line 54: @lucabello This fixture needs to be updated regardless for local testing. Why do you think having one fixture for the 22_04 and 24_04 packed charms is a bad thing? You are just hiding the hackery inside the fixture versus being explicit.

tests/integration/conftest.py
```diff
@@ -47,15 +47,21 @@ async def charm(ops_test: OpsTest) -> str:
...
 
 
+@pytest.fixture(scope="module")
+@timed_memoizer
+async def charm_22_04(charm) -> str:
+    """Charm (platform = ubuntu@22.04) used for integration testing."""
+    # Note: Use '22.04' in integration tests with Zookeeper, because that's Zookeeper's base
+    return charm.replace("24.04", "22.04")
```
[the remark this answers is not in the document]
MichaelThamm: I disagree, pleaes see [my comment below](https://github.com/canonical/opentelemetry-collector-operator/pull/34#discussion_r2226629064)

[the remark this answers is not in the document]
lucabello: Continuing the conversation there and resolving this thread

tests/integration/conftest.py
lucabello: Because IMHO the test itself shouldn't necessarily be concerned with the base - we're leaking an abstraction into the test implementation, while before the base concerns were separated from the test.

It's hackery either way, but I'm looking at what's easier to replace: what if CI passed all packed charms in a comma-separated list? What if (in a future where we're using spread) we're only building for one base and testing for it?

The current hackery is more self-contained, which is why I prefer it - but honestly I don't want to hold this up on some technical debt, so feel free to resolve this conversation and merge

Issue #27, opened 2025-07-17
Error state when removing relation to a principal or refreshing


lucabello:
### Bug Description

Currently, removing relation to a principal puts the charm into error state; this also happens on `upgrade`.

### To Reproduce

1. Deploy the vm charm and relate it to zookeeper
2. Either remove the relation to zookeeper, or try to refresh otelcol from a locally-packed version

### Environment

-

### Relevant log output

```shell
-
```

MichaelThamm:
## Requirements
1. An itest for upgrade
2. An itest for relating multiple subordinate otelcols to the same principal

## Testing
I tried

```bash
jd zookeeper zoo --trust
charmcraft pack
jd ./opentelemetry-collector_ubuntu@22.04-amd64.charm otel --trust
jrel zoo otel
# active/idle
jrmrel zoo otel
# active/idle for some time and then error
```

```shell
unit-otel-0: 16:07:46.621 DEBUG unit.otel/0.juju-log Processed dashboard '/var/lib/juju/agents/unit-otel-0/charm/grafana_dashboards/overview-dashboard.json': kept original uid '3031fb6fc42cebfa9502c30aee9969cb2c0af01e'
unit-otel-0: 16:07:52.429 ERROR unit.otel/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-otel-0/charm/lib/charms/operator_libs_linux/v2/snap.py", line 387, in _snap_daemons
    return subprocess.run(args, text=True, check=True, capture_output=True)
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['snap', 'restart', 'node-exporter']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-otel-0/charm/src/charm.py", line 448, in <module>
    ops.main(OpenTelemetryCollectorCharm)
  File "/var/lib/juju/agents/unit-otel-0/charm/venv/lib/python3.10/site-packages/ops/__init__.py", line 356, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-otel-0/charm/venv/lib/python3.10/site-packages/ops/_main.py", line 494, in main
[...]

PR #42, opened 2025-07-29
fix: Remove hook for cos-agent creates Error state

MichaelThamm:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/27

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -1158,7 +1158,7 @@ class COSAgentRequirer(Object):
             if not relation.units:
                 continue
             unit = next(iter(relation.units))
-            if not (raw := relation.data[unit].get(CosAgentProviderUnitData.KEY)):
+            if not (raw := relation.data.get(unit, {}).get(CosAgentProviderUnitData.KEY, {})):
                 continue
             if not (provider_data := self._validated_provider_data(raw)):
                 continue
--- a/src/charm.py
+++ b/src/charm.py
@@ -130,11 +130,20 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         super().__init__(framework)
         if hook() == "install":  # FIXME: install is not enough, we also need upgrade
             self._install()
+        logger.warning(f"---Hook: {hook()}")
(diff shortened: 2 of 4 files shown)


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -1158,7 +1158,7 @@ def _remote_data(self) -> List[Tuple[CosAgentProviderUnitData, JujuTopology]]:
             if not relation.units:
                 continue
             unit = next(iter(relation.units))
-            if not (raw := relation.data[unit].get(CosAgentProviderUnitData.KEY)):
+            if not (raw := relation.data.get(unit, {}).get(CosAgentProviderUnitData.KEY, {})):
```
MichaelThamm, line 1161: This was needed to fix the `KeyError: <ops.model.Unit zookeeper/0>` since Otelcol is no longer a subordinate to Zookeeper. I am not sure how this worked for grafana-agent?

lucabello: Likely this function is not called on every hook in grafana agent?

src/charm.py
```diff
@@ -130,11 +130,20 @@ def __init__(self, framework: ops.Framework):
...
         if hook() == "install":  # FIXME: install is not enough, we also need upgrade
             self._install()
+        logger.warning(f"---Hook: {hook()}")
         if hook() == "stop":
             self._stop()
+
         self._reconcile()
 
     def _reconcile(self):
+        if hook() in ["stop", "remove"]:
```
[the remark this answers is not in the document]
lucabello: Instead of doing this if here, do it at the start of the constructor (next to the other `if`s, right before `if hook() == "install"`, and return right there; the reconcile function should be event-independent by definition

pyproject.toml
```diff
@@ -6,7 +6,7 @@ version = "0.0"
 requires-python = "~=3.8"
 
 dependencies = [
-  "ops[tracing]>=2.17",
```
MichaelThamm, line 9: We decided to bump this to major version 3 in the k8s charm, we should have a tandem PR for this in the machine charm

PR #53, opened 2025-08-05
fix: save hash file

PietroPasotti:
## Issue
The hash file that we're using to determine if otelcol should be restarted is never updated.

## Solution
When the hash changes, update the file.


cbartz:
@PietroPasotti why didn't you merge the PR :smile:  Could have saved some headache. 

This was done in https://github.com/canonical/opentelemetry-collector-operator/pull/147 , I think this PR can be closed now.

PR #76, opened 2025-08-26
fix: Charm removes config file

MichaelThamm:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/77
There are 2 issues at hand:
1. The charm leaves old config files around.
2. The charm errors when we receive a `peers_relation_*` hook after `stop`. 

### 1. Cleanup up config files/directory
Rather than doing a reconcile on the config file dir, we remove the config on removal of the unit. This can mean that the directory can get out of sync in case the charm panics and cannot remove a config file. Since a stale config file gets merged in with the rest, it may be better to reconcile removal instead of removing when the unit departs. Reconciling could be a future issue since the PR solves most cases.

### 2. Cleanup on remove hook
Since the [juju remove-hook](https://documentation.ubuntu.com/juju/3.6/reference/hook/#remove) will always be the last hook executed on the charm, we should run our cleanup in there, i.e. snap and config file removal.

> If the unit has any relations active or any storage attached at the time the removal occurs, these will be cleaned up (in no specific order) between stop and remove. This means the unit will receive stop -> (*-relation-broken | *-storage-detaching) -> remove.

This means that we cannot remove the snaps on the `stop` hook otherwise, we will run into errors when we reconcile on a hook after the snaps were removed.

> [!WARNING]
> This goes against Juju's recommendation, "the stop event is emitted prior to this, and all necessary steps for handling removal should be handled there."

Solution:
1. Avoid reconciling on `remove` hooks, but still reconcile on all others
2. Remove the config file if other units exist, otherwise remove directory

## Testing Instructions

```shell
KEEP_MODELS=true tox -e integration -- tests/integration/test_removal_hooks.py
jssh ubuntu/0 'bash'
# Check for the removal of `/etc/otelcol/config.d/*.yaml` when scaling down a unit
# Check for the existence of `/etc/otelcol/config.d` when removing relation
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -465,6 +465,7 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
             self.snap(snap_name).ensure(state=snap.SnapState.Absent)
         except (snap.SnapError, SnapSpecError) as e:
             raise SnapInstallError(f"Failed to uninstall {snap_name}") from e
+        logger.info(f"{snap_name} snap was uninstalled")
 
     def _remove_node_exporter(self):
         """Coordinate node-exporter snap removal."""
@@ -474,7 +475,6 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         manager.unregister(snap_name, snap_revision)
         if not manager.is_used_by_other_units(snap_name):
             self._remove_snap(snap_name)
-
 
     def _remove_opentelemetry_collector(self):
         """Coordinate opentelemetry-collector snap and config file removal."""
@@ -486,12 +486,12 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
             config_filename = f"{SnapRegistrationFile._normalize_name(self.unit.name)}.yaml"
             config_path = LocalPath(os.path.join(CONFIG_FOLDER, config_filename))
             config_path.unlink()
-            logger.info(f"Removed the opentelemetry-collector config file: {config_path}")
+            logger.info(f"removed the opentelemetry-collector config file: {config_path}")
             self.snap("opentelemetry-collector").restart()
         else:
             self._remove_snap(snap_name)
             shutil.rmtree(LocalPath(CONFIG_FOLDER))
-            logger.info(f"Removed the opentelemetry-collector config folder: {CONFIG_FOLDER}")
+            logger.info(f"removed the opentelemetry-collector config folder: {CONFIG_FOLDER}")
 
         # TODO: Luca if the snap is used by other units, we should probably `ensure`
         # that the max_revision is installed instead.


src/charm.py
```diff
@@ -455,50 +457,37 @@ def _install_snaps(self) -> None:
...
+                    self.snap(snap_name).ensure(state=snap.SnapState.Absent)
+                except (snap.SnapError, SnapSpecError) as e:
+                    raise SnapInstallError(f"Failed to uninstall {snap_name}") from e
+                if snap_name == "opentelemetry-collector":
+                    # Remove the config dir
+                    shutil.rmtree(LocalPath(CONFIG_FOLDER))
+                    logger.info(
+                        f"Removed the opentelemetry-collector config folder: {CONFIG_FOLDER}"
+                    )
+
```
sed-i, line 488: After the if-else block, do we need to re-merge the config files / restart the snap?

MichaelThamm: This is a great point, but I think we would only want to do it in the IF block because this is where we know there are other units registered. In the ELSE, we remove the snap, so no need to restart/merge.

src/charm.py
```diff
@@ -455,50 +457,37 @@ def _install_snaps(self) -> None:
...
-            #     f.flush()
-            pass
-
-    def _stop(self) -> bool:
-        """Coordinate snap and config file removal.
-
-        If the snap is solely used by this unit, then skip reconciling the charm since it depends
-        on snap operations.
-        """
+    def _remove(self):
```
Abuelodelanada, line 460: What do you think about a refactor like this one?

- One indentation level removed
- Guard clause added: if the snap_name is not `opentelemetry-collector` we `continue`
- The removal of `CONFIG_FOLDER` is done in the same `try...except` the snap is removed.

```suggestion
    def _remove(self):
        """Coordinate snap and config file removal."""
        manager = SingletonSnapManager(self.unit.name)

        for snap_name in SnapMap.snaps():
            if snap_name != "opentelemetry-collector":
                continue

            snap_revision = SnapMap.get_revision(snap_name)
            manager.unregister(snap_name, snap_revision)

            if manager.is_used_by_other_units(snap_name):
                # Remove the config file for the unit
                config_filename = f"{SnapRegistrationFile._normalize_name(self.unit.name)}.yaml"
                config_path = LocalPath(os.path.join(CONFIG_FOLDER, config_filename))
                config_path.unlink()
                logger.info(f"Removed the opentelemetry-collector config file: {config_path}")
                continue

            # Remove the snap
            self.unit.status = MaintenanceStatus(f"Uninstalling {snap_name} snap")
            try:
                self.snap(snap_name).ensure(state=snap.SnapState.Absent)
                logger.info(f"{snap_name} snap uninstalled")
                # Remove the config dir
                shutil.rmtree(LocalPath(CONFIG_FOLDER))
                logger.info(f"removed the opentelemetry-collector config folder: {CONFIG_FOLDER}")
            except (snap.SnapError, SnapSpecError) as e:
                raise SnapInstallError(f"Failed to uninstall {snap_name}") from e
```

PR #90, opened 2025-09-10
Set workload version

sed-i:
## Issue
Currently there is no version indicated in juju status.

## Solution
Set workload version as part of the reconciler.
- Do not catch `[CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError)` - if the command changes we want it to fail loudly in CI
- It could be handy to clear the workload version on stop, but in the reconciler we would need to do it on all teardown events (peer relation departed, stop, remove) to prevent the workload version from "flickering". Not worth it so not doing it :)

In tandem with https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/125.


lucabello (approved):

sed-i:
I see @lucabello is working on improving the itests in #91 so merging here on the flaky red.

PR #91, opened 2025-09-10
feat: improve integration tests and path_exclude

### LINKED ISSUE #56: Improve itengration tests
### Enhancement Proposal

In our integration tests we are relying on otelcol internal telemetry to check for logs or metrics arriving the collector.

This is far from ideal, because we have to configure the charm differently from the default settings in order to [obtain those logs](https://github.com/canonical/opentelemetry-collector-operator/blob/ed9f138c64a2ab7d02fe19abf74e13d58b0bb80f/tests/integration/test_cos_agent.py#L19).

lucabello:
This PR does two things:
1. fix integration tests by making sure we're not starving a host during tests
2. fix a bug in the `path_exclude` functionality

Number (1) is done by configuring the charm to not scrape `syslog` and other files which would cause a recursive feedback loop (otelcol scrapes syslog, it emits a log about that to syslog, then scrapes syslog again, infinitely). This is achieved via the already-existing `path_exclude` config option, which allows to ignore some files from the log scrape job.

Number (2) is a direct consequence of (1), as I realized the `path_exclude` functionality had a bug: we were splitting globs on commas, which prevents us from using a certain type of glob:

> For example, `/var/log/**/{app_one,app_two}.log` will result in the agent ignoring both `/var/log/app_one.log` and `/var/log/app_two.log`.

I changed the path delimiter to be `;`, just like in the `PATH` environment variable, to allow for these patterns.

Closes #56 

---

Additional things:
- `fetch-lib`


MichaelThamm:
Remember to remove the [verbosity setting in the conftest fixture ](https://github.com/canonical/opentelemetry-collector-operator/blob/d5da6f866c2e08561c7ba70a68396a2e2ba68430/tests/integration/conftest.py#L66) and so we do not override the [setting in config_builder](https://github.com/canonical/opentelemetry-collector-operator/blob/d5da6f866c2e08561c7ba70a68396a2e2ba68430/src/config_builder.py#L175) :)

PR #128, opened 2025-11-17
feat: add ability to scrape log files by name

### LINKED ISSUE #120: Add ability to scrape log files by name (without log slot)
Via @weiiwang01 @cbartz @javierdelapuente @arturo-seijas

Charms related over cos-agent should be able to provide a list of files to scrape.
This is useful because then specific files from `/var/log` would have topology labels associated with the principal.
(Currently logs come into loki with otelcol's topology.)

https://github.com/canonical/opentelemetry-collector-operator/blob/02b7fe0135d1d6e81ea27f4b6106185f2684cd60/src/charm.py#L326

However, this may pose a problem for logs coming from juju-info relation, because there is no charm_type label.

Also, since otelcol is a strictly confined snap, it will only have access to files matching declared (and connected) interfaces.

arturo-seijas:
## Issue

Fixes #120


MichaelThamm:
Related to this PR:
- https://github.com/canonical/grafana-agent-operator/pull/363

PR #130, opened 2025-11-20
chore: implement new prometheus remote write requirements

sinapah:
## Issue

Tracking issue with further context: https://github.com/canonical/cos-lib/issues/167

## Solution

1. Bumps `cosl` so new generic alert rules (with new descriptions) are used.
2. Fetches charm libs so new changes from Prometheus remote write lib are present (lib patch 11).
3. Passes the name of this charm's peer relation when instantiating `PromtheusRemoteWriteConsumer`.

## Testing Instructions

1. Pack this charm and relate it to Ubuntu and Prometheus.
2. Scale up Ubuntu to 2.
3. On the Prometheus alerts page, you should now see 2 HostMetricsMissing alerts for Otelcol, one per each unit.
image

4. If you `sudo snap stop opentelemetry-collector` in one of the two LXD machines, one of the two HostMetricsMissing alerts should be triggered AND the AggregatorMetricsMissing alert should NOT be triggered.
image

5. And if you stop the snap in both LXD machines, both alerts will be triggered and since ALL units are down, the AggregatorMetricsMissing alert now should fire too.

image

The severity of the HostMetricsMissing alerts must be critical since Otelcol machine is a subordinate charm.
image

### DIFF
--- a/lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
@@ -23,11 +23,11 @@ import socket
 import subprocess
 import tempfile
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 
 import yaml
 from cosl import JujuTopology
-from cosl.rules import AlertRules, generic_alert_groups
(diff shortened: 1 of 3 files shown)


MichaelThamm (changes requested):
What is the reason you are seeing the `HostMetricsMissing` alert for `otel/4` in a Pending state? Can you please clarify in the PR description AND/OR update the image to show both alerts as no firing. Also, can you please add the results for when you take one of the units down with `snap stop otelcol` and show the alert rules, including the `AggregatorMetricsMissing` rule.

Issue #104, opened 2025-10-17
Log messages fill up disk during Loki outage


alexdlukens:
### Bug Description

We have experienced an outage on the IS COS instance in PS6. During the outage, otel collector fails to send logs to loki. These errors are logged as tracebacks in `/var/log/syslog`, and quickly fill up the disk (see logs below).  Unclear the  exact time it took to fill the disk (40G), but the outage has been for < 12h.

I *think* this is a charm bug, due to how debug exporters are configured by default in the code https://github.com/canonical/opentelemetry-collector-operator/blob/af97af4d83c1fd4b20378534466f5a50472c04e1/src/config_builder.py#L256-L271.

Could the configuration be tweaked here to reduce the logging? Increasing the sampling time, disable internal logger, set verbosity to "basic"?

### To Reproduce

Relate OTEL collector to Loki, then disrupt network connection.

### Environment

This is the `stg-discourse-db-ps6` environment on `PS6`
charm version: `0.130.0` on `2/edge` channel, `rev78`
integrated with loki via: `loki-push-api:logging <-> opentelemetry-collector:send-loki-logs`

### Relevant log output

```shell
Oct 17 13:45:15 juju-f1ec08-stg-discourse-db-ps6-2 opentelemetry-collector.opentelemetry-collector[2334218]: 2025-10-17T13:45:15.733Z#011error#011internal/queue_sender.go:51#011Exporting failed. Dropping data.#011{"resource": {"service.instance.id": "cc7818c6-ca56-4ca9-be1a-742e364c5de9", "service.name": "otelcol", "service.version": "0.130.0"}, "otelcol.component.id": "loki/send-loki-logs/0", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "error": "not retryable error: Permanent error: HTTP 404 \"Not Found\": 404 page not found", "dropped_items": 100}
Oct 17 13:45:15 juju-f1ec08-stg-discourse-db-ps6-2 opentelemetry-collector.opentelemetry-collector[2334218]: go.opentelemetry.io/collector/exporter/exporterhelper/internal.NewQueueSender.func1
Oct 17 13:45:15 juju-f1ec08-stg-discourse-db-ps6-2 opentelemetry-collector.opentelemetry-collector[2334218]: #011go.opentelemetry.io/collector/exporter@v0.130.0/exporterhelper/internal/queue_sender.go:51
[...]

PR #138, opened 2025-11-28
fix!: Otelcol self-mon logs syslog explosion

MichaelThamm:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/104

CI is fixed by:
- https://github.com/canonical/opentelemetry-collector-operator/pull/162

Related:
- https://github.com/canonical/opentelemetry-collector-operator/pull/142/
- https://github.com/canonical/opentelemetry-collector-operator/issues/140

## Solution

This solution comes in 2 parts:
1. Reducing the rate of explosion
2. Stopping the explosion

### Reducing the rate of explosion
After [some investigation](https://github.com/canonical/opentelemetry-collector-operator/issues/104#issuecomment-3587386054), the solution to fix the `syslog` explosion is multifaceted:
1. The internal logging stacktrace can be disabled (`disable_stacktrace`) to reduce error log lines
2. The logging level is already set to `WARN` which is good
```yaml
  telemetry:
    logs:
      disable_stacktrace: true
      level: WARN
```
3. [Configure](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#retry-on-failure) `retry-on-failure` to retry less aggressively.
	- https://github.com/canonical/opentelemetry-collector-operator/blob/e611799973e89ebaf12e3d792e67a09d76a1f48e/src/config_manager.py#L153

> [!NOTE]
> This was not implemented in this PR because the other solutions should resolve the issue 

### Stopping the explosion
See the `Context` section below for the edge case that creates this problem. Since this is only an issue when the log server (exporter endpoint) fails (e.g. stops, firewall, etc.), the Otelcol `>=WARN` logs can be written to a designated file to not interfere with our [`/var/log` file receiver](https://github.com/canonical/opentelemetry-collector-operator/blob/e611799973e89ebaf12e3d792e67a09d76a1f48e/src/charm.py#L332)
```yaml
  telemetry:
    logs:
      output_paths:
        - /var/snap/opentelemetry-collector/common/otelcol.log
```
This means that `sudo snap logs -f opentelemetry-collector` is no longer useful and instead we would have to ask issue reporters to attached the output of the `otelcol.log` file instead.

> [!WARNING]
> The `output_paths` log files will not be automatically rotated, unlike how syslog is rotated:
> https://github.com/canonical/opentelemetry-collector-operator/pull/138#issuecomment-3598176938

### The true problem
There exists a recursion problem IFF the log exporter endpoint goes down. This will generate `Exporting failed` logs which otelcol attempts to write to Loki again, and failing. This goes on infinitely at the [configure](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#retry-on-failure) `retry-on-failure`. 
image

### Logrotate
There is an upstream initiative to handle logrotation of otelcol log files:
- https://github.com/open-telemetry/opentelemetry-collector/issues/10768

The manpage for `logrotate`:
- https://linux.die.net/man/8/logrotate

## Testing Instructions

- [x] Test the alert rules by stopping Loki with pebble
- [x] https://github.com/canonical/opentelemetry-collector-operator/pull/138#issuecomment-3598176938

## Upgrade Notes

- Otelcol logs are in `/var/snap/opentelemetry-collector/common/otelcol.log` instead of journal.
- New `logrotated` dependency

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -197,7 +197,14 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         # Refresh system certs
         # This must be run after receive_ca_cert and/or receive_server_cert because they update
         # certs in the /usr/local/share/ca-certificates directory
-        refresh_certs()
+        # Only refresh certs when they actually change (upgrade-charm or cert relation changes)
+        current_hook = hook()
+        if current_hook in (
+            "upgrade-charm",
+            "receive_ca_cert-relation-changed",
+            "receive_server_cert-relation-changed",
+        ):
+            refresh_certs()
 
         # Global scrape configs
         global_configs = {
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -10,6 +10,7 @@ import yaml
 from tenacity import retry, stop_after_attempt, wait_fixed
 
 
+# Exclude some logs to avoid circular ingestion during tests
 PATH_EXCLUDE: Final[str] = "/var/log/**/{cloud-init-output.log,syslog,auth.log};/var/log/juju/**"
 # Configure debug exporters for all pipelines to inspect / assert against the OTLP data
 ENABLE_BASIC_DEBUG_EXPORTERS: Final[Dict[str, str]] = {
--- a/tests/unit/test_refresh_certs.py
+++ b/tests/unit/test_refresh_certs.py
@@ -1,0 +1,57 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""Tests for refresh_certs behavior."""
+
+from unittest.mock import patch
+
+import pytest
+from ops.testing import Relation, State
+
+
+@pytest.mark.parametrize(
+    "relation_name,remote_app_data",
+    [
+        ("receive-ca-cert", {"certificates": "[]"}),
+        ("receive-server-cert", {}),
+    ],
+)
+def test_refresh_certs_called_on_cert_relation_changed(ctx, relation_name, remote_app_data):
+    """Test that refresh_certs is called on certificate relation-changed hooks."""
+    # GIVEN a certificate relation exists
+    cert_relation = Relation(relation_name, remote_app_data=remote_app_data)
+    state = State(
+        leader=True,
+        relations=[cert_relation],
+    )
+
+    # Mock refresh_certs to track if it's called
+    with patch("charm.refresh_certs") as mock_refresh_certs, \
+         patch("integrations._add_alerts"):
+        # WHEN the relation changed event is executed
+        ctx.run(ctx.on.relation_changed(cert_relation), state)
+
+        # THEN refresh_certs should be called
+        mock_refresh_certs.assert_called_once()
+
+
+@pytest.mark.parametrize(
+    "hook_name,event_method",
+    [
+        ("update-status", "update_status"),
+        ("config-changed", "config_changed"),
+    ],
+)
+def test_refresh_certs_not_called_on_non_cert_hooks(ctx, hook_name, event_method):
+    """Test that refresh_certs is NOT called on non-certificate hooks."""
+    # GIVEN a non-certificate hook is being executed
+    state = State(leader=True)
+
+    # Mock refresh_certs to track if it's called
+    with patch("charm.refresh_certs") as mock_refresh_certs, \
+         patch("integrations._add_alerts"):
+        # WHEN the hook is executed
+        event = getattr(ctx.on, event_method)()
+        ctx.run(event, state)
+
+        # THEN refresh_certs should NOT be called
+        mock_refresh_certs.assert_not_called()


MichaelThamm:
- Likely, the `output_paths` log files will not be rotated, unlike how syslog is rotated.
- If we use `output_paths` then `sudo snap logs -f opentelemetry-collector` is no longer really useful and instead we would have to ask issue reporters to attached the output of the otelcol.log file instead.

tests/integration/test_log_rotation.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+
+async def test_log_rotation(juju: jubilant.Juju):
+    # GIVEN the log file is present on disk
+    files = juju.ssh("otelcol/0", f"ls {LOG_DIR}").strip().split("  ")
+    assert files == ["otelcol.log"]
+
+    # WHEN the log rotation is run manually
+    juju.ssh("otelcol/0", "sudo logrotate -f /etc/logrotate.d/otelcol").strip()
```
MichaelThamm: Does this work just by adding it to the `logrotate.d` dir?

src/logrotate.d/otelcol
```diff
@@ -0,0 +1,15 @@
+/var/snap/opentelemetry-collector/common/otelcol.log
+{
+    rotate 4
+    weekly
```
MichaelThamm: This was copied from the `rsyslog` default:
```
rotate 4  # 4 rotations before logs start getting dropped
weekly  # rotate weekly
```

There is another param named `size` which can be set like so:
```
size 100k
```

This seems to be a balancing game of large deployments (many logs) vs. smaller ones. If we do not add a `size` param, these files may grow a lot, although we now have the logging level set at `WARN` which will reduce the logs.

tests/integration/test_log_rotation.py
```diff
@@ -0,0 +1,53 @@
...
+from helpers import PATH_EXCLUDE
+
+from constants import INTERNAL_TELEMETRY_LOG_FILE
+
+LOG_DIR = str(pathlib.Path(INTERNAL_TELEMETRY_LOG_FILE).parent)
+
+
+async def test_deploy(juju: jubilant.Juju, charm: str):
+    # GIVEN an OpenTelemetry Collector charm and a principal
+    ## NOTE: /var/log/cloud-init.log and /var/log/cloud-init-output.log are always present
```
MichaelThamm:
```suggestion
```

tests/integration/test_log_rotation.py
```diff
@@ -0,0 +1,53 @@
...
+    juju.deploy(charm, app="otelcol", config={"path_exclude": PATH_EXCLUDE})
```
MichaelThamm: We don't need `PATH_EXCLUDE` anymore since we do not have the recursion problem. This would filter out irrelevant logs in the test, but adds clutter to the test.
```suggestion
    juju.deploy(charm, app="otelcol")
```

tests/integration/test_log_rotation.py
```diff
@@ -0,0 +1,53 @@
...
+    juju.wait(
+        lambda status: jubilant.all_blocked(status, "otelcol"),
+        error=jubilant.any_error,
+        timeout=420,
+    )
+
+
+async def test_log_rotation(juju: jubilant.Juju):
+    # GIVEN the log file is present and is configured for log rotation
+    breakpoint()
```
Abuelodelanada: Do not forget to remove this:


```suggestion
```

src/logrotate.d/otelcol
```diff
@@ -0,0 +1,15 @@
...
+    notifempty
+    compress
+    delaycompress
+    sharedscripts
+    postrotate
+        if [ -d /run/systemd/system ]; then
+            kill -HUP $(pidof otelcol)
+        fi
+    endscript
+}
```
sed-i: Do we have an analogous problem in the k8s charm?

src/logrotate.d/otelcol
```diff
@@ -0,0 +1,15 @@
...
+    missingok
+    notifempty
+    compress
+    delaycompress
+    sharedscripts
+    postrotate
+        if [ -d /run/systemd/system ]; then
+            kill -HUP $(pidof otelcol)
+        fi
+    endscript
```
sed-i: Can we have comments in this file? Explain the purpose briefly?

src/charm.py
```diff
@@ -594,6 +601,25 @@ def _configure_node_exporter_collectors(self):
         ne_snap = self.snap("node-exporter")
         self._set_snap_configs_with_retry(ne_snap, configs)
 
+    def _configure_logrotate(self, config: str):
```
sed-i: config arg seems unused. Which seems correct.

src/charm.py
```diff
@@ -212,6 +214,9 @@ def _reconcile(self):
             max_elapsed_time_min=cast(int, self.config.get("max_elapsed_time_min")),
         )
 
+        # Self-mon logging
+        self._configure_logrotate(config_manager.config.build())
```
sed-i: At first glance, logrotate config shouldn't depend on otelcol config.

tests/integration/test_log_rotation.py
```diff
@@ -0,0 +1,49 @@
...
+        timeout=420,
+    )
+    juju.wait(
+        lambda status: jubilant.all_blocked(status, "otelcol"),
+        error=jubilant.any_error,
+        timeout=420,
+    )
+
+
+async def test_log_rotation(juju: jubilant.Juju):
```
sed-i: This test seems brittle and not testing charm features.
Maybe something like:
```
# GIVEN a fresh charm
# THEN there is a single otelcol log file
# WHEN VM system clock is bumped by (week / 4) days
# THEN there the otelcol log file is rotated
```
?

src/charm.py
MichaelThamm: I was considering using the config arg to check if `service:telemetry:logs:output_paths` is configured. If so, only then do we configure logrotate. I moved this check to a unit test instead.

sinapah (approved):
Loki, Grafana, Tempo, Mimir!

PR #147, opened 2025-12-08
fix: Fix constant otel_collector restart on update-status

### LINKED ISSUE #146: otelcol constantly restarts because /opt/otelcol_reload is never persisted
### Bug Description

The charm constantly restarts the opentelemetry-collector snap due to a logic bug in the config/cert hashing mechanism. In src/charm.py, the logic that determines when to restart the snap depends on the presence and contents of the file /opt/otelcol_reload. If that file does not exist, the charm always interprets the configuration as changed, so the restart block will execute every hook firing (commonly every 5 minutes during update-status) leading to constant restarts. There is currently no code which persists (writes) the hash value to /opt/otelcol_reload after a restart, so the restarts never stop.

### To Reproduce

1. Deploy the opentelemetry-collector-operator charm on any model/environment
2. Observe unit syslogs or juju debug-log
3. Notice that the opentelemetry-collector service is stopped and started (restarted) every 5 minutes, even with no config or relation changes
4. Check the unit and see that /opt/otelcol_reload does not exist after charm operations
5. Confirm the restart block in src/charm.py - existence of the file prevents restart, but it is never written to disk.

### Environment

Juju model: 3.6
Cloud: OpenStack
Charm version: code is in current main branch but we tested with rev 91
Track/channel: 2/stable

### Relevant log output

```shell
Excerpt from syslog:

Dec  7 00:08:15 systemd[1]: Stopping Service for snap application opentelemetry-collector.opentelemetry-collector...
Dec  7 00:08:15 systemd[1]: Stopped Service for snap application opentelemetry-collector.opentelemetry-collector.
Dec  7 00:08:15 systemd[1]: Started Service for snap application opentelemetry-collector.opentelemetry-collector.

This repeats every 5 minutes, aligned with Juju's update-status hook frequency.
```

### Additional context

Proposed fix: After the restart logic in charm.py, add a line to persist the new hash value to /opt/otelcol_reload – e.g. hash_file.write_text(current_hash). This will prevent redundant restarts unless configuration or certs really change.

cbartz:
## Issue

Fix #146 

## Solution

Persist the hash

## Testing Instructions

1. Deploy the charm.
2. Configure update-status interval to a short period.
3. Observe that otel collector (systemd service) is not restarted on every update status.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -461,6 +461,7 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         if current_hash != old_hash:
             for snap_name in SnapMap.snaps():
                 self._restart_snap(self.snap(snap_name))
+            hash_file.write_text(current_hash)
 
         # Set status
         if self._has_server_cert_relation and not is_tls_ready():


Abuelodelanada (approved):
Good catch! Thanks!

PR #156, opened 2025-12-10
fix: config hash path

### LINKED ISSUE #155: Ensure the reload hash considers the existence of multiple units
A recent PR (https://github.com/canonical/opentelemetry-collector-operator/issues/146) ensured that the config hash for Otelcol machine is written to disk and upon a change in that hash, a restart is supposed to happen.

We need to ensure that we slightly modify the logic so that a hash is written to disk per each Otelcol unit on that machine. For example, if, using the singleton pattern, we may have multiple Otelcols on the same machine. At present, they will both try to store their config hash in a file called `otelcol_reload`. We should ensure that they write a file which has a unique name based on their unit names e.g. `otelcol_reload_otel1/0` and `otelcol_reload_otel2/0`.

sinapah:
## Issue

Fixes #155.
Todo
- [ ] Backport into track 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,14 +14,14 @@ $ juju deploy opentelemetry-collector
 ```
 The `opentelemetry-collector` operator for machines is a subordinate charm.
 
-It needs to be related to a principal charm. The principal charm needs to support relations through the `cost-agent` or `juju-info` interfaces.
+It needs to be related to a principal charm. The principal charm needs to support relations through the `cos-agent` or `juju-info` interfaces.
 For example, if you have the `ubuntu` principal, you can relate it to this charm:
 
 ```bash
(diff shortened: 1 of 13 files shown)


sed-i (approved):
Do we need this for the k8s charm?

sinapah:
@sed-i 
It shouldn't be necessary for K8s. In that charm, Pebble will restart automatically if the config file or any cert files change.
https://github.com/canonical/opentelemetry-collector-k8s-operator/blob/e12107c2675889979aeac519fc9eeefffce4c838/src/charm.py#L255-L265

sinapah:
Closing this PR as #162 had to implement the same fixes to pass CI. This PR is not unnecessary.

PR #168, opened 2026-01-19
feat: Use the cos-lib reconciler

MichaelThamm:
## Issue

We added a reconciler, and later found out that there is [a better way to do it](https://github.com/canonical/cos-lib/blob/main/src/cosl/reconciler.py). This was needed for the tandem PR in the k8s charm and is being added to this charm for parity:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/173

## Solution

Implement the `cosl` reconciler in the charm's constructor.

I added the `ops.charm.ActionEvent` as a reconcilable event in `charm.py`:
```python
observe_events(self, all_events, self._reconcile)
observe_events(self, [ops.charm.ActionEvent], self._reconcile)
```

I thought maybe this would be a reasonable event to add as a feature to cosl, but I think I will close this PR:
- https://github.com/canonical/cos-lib/pull/174

### Checklist


MichaelThamm:
The `LokiPushApiProvider` lib was not designed to handle the reconciler-like pattern when using `cosl.reconciler`

PR #181, opened 2026-02-12
fix: add dynamic port allocation for node-exporter

### LINKED ISSUE #176: Can not add relation with landscape-server due to port conflict with node-exporter
### Bug Description

When adding a relation between landscape-server and opentelemetry-collector, landscape-server charm gets in a blocked state because of a conflict in listening port between node-exporter (9100/tcp) and landscape's package-upload (9100/tcp).
Since both, landscaper-server and opentelemetry-collector charms do not have a config option to change the listening port, at the moment there is no workaround for this.
I have raised a similar issue for the landscape-server charm as well.
- https://bugs.launchpad.net/landscape-charm/+bug/2139659

However, since opentelemetry-collector is the subordinate charm to monitor the application/workload, I think adding a config option to change the port on the landscape-server charm is a bit too excessive since it could happen on other applications as well. 

A config option to change the listen port would help to avoid conflict with other application charms.

### To Reproduce

1. juju deploy landscape-scalable
2. juju deploy opentelemetry-collector --base ubuntu@22.04
3. juju integrate landscape-server opentelemetry-collector

### Environment

1. opentelemetry-collector charm from latest/stable
2. juju v.3.6

### Relevant log output

_No reponse_

### Additional context

_No response_

Abuelodelanada:
## Issue

This PR fixes #176 
Tandem PR: https://github.com/canonical/node-exporter-snap/pull/24

When `otelcol` charm is deployed, and hence `node-exporter` snap is installed in the host, we may find a situation in which `node-exporter` try to use the `9100` port that is already used

## Solution

Implemented dynamic port allocation for `node-exporter` configuration:

- Added a public `find_available_port()` function that searches for available port starting from a configurable default
- Modified `_configure_node_exporter()` to dynamically detect and use the first available port instead of a hardcoded value

### Checklist

## Context

Node-exporter is a Prometheus exporter that collects hardware and OS metrics. When deployed as a subordinate charm, multiple instances can run on the same physical/virtual 
chine. Each instance needs its own unique port to expose metrics without conflicts.

The solution uses Python's `socket` library to attempt binding to ports sequentially. If a port is already in use (raises `OSError`), the function moves to the next port 
til it finds an available one.

### Unit Tests

Run the unit tests to verify port allocation logic:

```bash
tox -e unit -- tests/unit/test_node_exporter_config.py -v
```

This tests three scenarios:

- Default port 9100 is available
- Port 9100 is occupied (selects 9101)
- Multiple ports occupied (finds first available)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -182,6 +182,7 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         self._reconcile()
 
     def _reconcile(self):
+        self._node_exporter_port = find_available_port(start_port=NODE_EXPORTER_DEFAULT_PORT)
         insecure_skip_verify = cast(bool, self.config.get("tls_insecure_skip_verify"))
         topology = JujuTopology.from_charm(self)
         # NOTE: Only the leader aggregates alerts, to prevent duplication. COS Agent alerts
@@ -292,7 +293,7 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
                             "static_configs": [
                                 {
                                     "targets": [
-                                        "0.0.0.0:9100"  # TODO: extract this node-exporter port somewhere
+                                        f"0.0.0.0:{self._node_exporter_port}"
                                     ],
                                     "labels": {
                                         "instance": socket.getfqdn(),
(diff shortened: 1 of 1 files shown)


MichaelThamm:
You also have a failing CodeQL test:
- https://github.com/canonical/opentelemetry-collector-operator/pull/181/checks?check_run_id=63536635946

PR #235, opened 2026-03-31
feat: Drop limit: 1 in send-traces

mmkay:
## Issue

The send-traces endpoint had an unnecessary limit: 1. It isn't because of any reasonable restrictions: opentelemetry-collector can send traces to more than one backend.

## Solution

Drop limit: 1 and make sure all the backends get the traces.

### Checklist

## Context

We did not touch the send-charm-traces endpoint because it's responsible for sending otel-collector's own charm traces. ops[tracing] does not allow for more than 1 destination for the charm traces so I think in these cases there'd be another otel collector with a custom configuration.

Although we will consider retiring the interface in future, given the near LTS, we want to make sure there's enough support for multiple tracing backends in the current charms.

Contributes to solving https://github.com/canonical/tempo-operators/issues/36 .

Tandem PR: https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/228

## Testing Instructions

The integration tests should cover the testing part.


Abuelodelanada (approved):
LGTM! Thanks!

src/charm.py
```diff
@@ -264,8 +264,10 @@ def _reconcile(self):
             sampling_rate_workload=cast(float, self.config.get("tracing_sampling_rate_workload")),
             sampling_rate_error=cast(float, self.config.get("tracing_sampling_rate_error")),
         )
-        if tracing_otlp_http_endpoint := integrations.send_traces(self):
-            config_manager.add_traces_forwarding(tracing_otlp_http_endpoint)
+        tracing_endpoints = integrations.send_traces(self)
+        for idx, endpoint in enumerate(tracing_endpoints):
```
sed-i: Nit: Is the order stable, e.g. sorted by rel-id?
It would be nice to be able to avoid unnecessary config reloads.

src/charm.py
```diff
@@ -264,8 +264,10 @@ def _reconcile(self):
...
+            config_manager.add_traces_forwarding(endpoint, identifier=str(idx))
```
sed-i: `idx` is opaque. Would it be easy enough to use rel-id for the identifier?

src/config_manager.py
```diff
@@ -501,21 +501,22 @@ def add_traces_processing(
...
-            Currently, only one endpoint is supported due to limitations in the
-            Tempo charm. The exporter will be added to the 'traces' pipeline.
+            identifier: A unique string used to disambiguate the exporter name
+                (e.g. the relation index). The exporter will be named
+                ``otlphttp/send-traces-{identifier}``.
         """
         self.config.add_component(
             Component.exporter,
-            "otlphttp/send-traces",
+            f"otlphttp/send-traces-{identifier}",
```
sed-i: Iirc, typically we'd "namespace" (not suffix) by rel-id

```suggestion
            f"otlphttp/rel-{rel_id}/send-traces",
```

src/integrations.py
```diff
@@ -361,9 +362,12 @@ def send_traces(charm: CharmBase) -> Optional[str]:
...
     charm.__setattr__("tracing", tracing_requirer)
-    if not tracing_requirer.is_ready():
-        return None
-    return tracing_requirer.get_endpoint("otlp_http")
+    return [
+        endpoint
+        for rel in charm.model.relations["send-traces"]
+        if tracing_requirer.is_ready(rel)
+        if (endpoint := tracing_requirer.get_endpoint("otlp_http", rel)) is not None
+    ]
```
sed-i: Should this be a [rel-id](https://github.com/canonical/operator/blob/df731e515a06617f500bce2724762cdcf47b9d51/testing/src/scenario/state.py#L520) mapping?

```suggestion
    return {
        rel.id: endpoint
        for rel in charm.model.relations["send-traces"]
        if tracing_requirer.is_ready(rel)
        if (endpoint := tracing_requirer.get_endpoint("otlp_http", rel)) is not None
    }
```

tests/integration/test_multi_backend_tracing.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    juju.wait(
+        lambda status: jubilant.all_agents_idle(
+            status, "otelcol", "otelcol-b1", "otelcol-b2", "ubuntu-main", "ubuntu-b1", "ubuntu-b2"
+        ),
+        timeout=120,
+    )
+
+
+def test_two_exporters_in_config(juju: jubilant.Juju):
```
sed-i: Should we move these assertions into a unit test?
It's nice if the brittle `"otlphttp/send-traces-"` literal is on a cheaper test.
Unless I'm missing the extra coverage an itest gives us here.

tests/integration/test_multi_backend_tracing.py
```diff
@@ -0,0 +1,106 @@
...
+
+def test_no_ambiguous_relation_error(juju: jubilant.Juju):
+    """No unit enters error state, which is what AmbiguousRelationUsageError would cause."""
+    # An unhandled AmbiguousRelationUsageError in a charm hook drives the unit into
+    # error state. Asserting clean status is more reliable than grepping log output,
+    # which is subject to timing and buffering.
+    assert not jubilant.any_error(juju.status())
+
+
+def test_relation_removal_reconfigures_cleanly(juju: jubilant.Juju):
```
sed-i: These assertions also seem like could be covered by a utest.

sed-i (commented):
Thanks @mmkay!
A couple questions.

src/charm.py
mmkay: I have added the sorting so that we're certain.

src/config_manager.py
mmkay: Good point, done.

src/integrations.py
mmkay: Yes, definitely!

tests/integration/test_multi_backend_tracing.py
mmkay: Indeed, I moved the assertions and removed the test altogether.

tests/integration/test_multi_backend_tracing.py
mmkay: Indeed, I moved the assertions and removed the test altogether.

src/charm.py
mmkay: It is possible - applied the changes.

PR #237, opened 2026-04-01
refactor: Remove asyncio from itests

MichaelThamm:
## Issue

We do not need asyncio in our itests since jubilant is synchronous.

## Solution

Remove asyncio and use pytest-jubilant instead of pytest-operator [like we do in our monorepos](https://github.com/canonical/loki-operators/blob/f47f3595f98a92afd166690b05fc2f85e252b8a0/tests/integration/conftest.py#L28).

### Checklist

## Testing Instructions

See CI

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -4,6 +4,7 @@
 """Conftest file for integration tests."""
 
 import functools
+import glob
 import logging
 
 import os
@@ -12,7 +13,7 @@ from datetime import datetime
 from pathlib import Path
 
 import pytest
-from pytest_operator.plugin import OpsTest
+import sh
 import jubilant
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


tests/integration/conftest.py
```diff
@@ -3,72 +3,62 @@
...
-    # if charm_file := os.environ.get("CHARM_PATH"):
-    #     return charm_file
-
-    charm = await ops_test.build_charm(".")
-    charm = str(charm).replace("24.04", "22.04")
-    assert charm
-    return charm
+    if channel_from_env := os.getenv(charm_channel_key):
+        charm = "opentelemetry-collector"
+        logger.info("Using published %s charm from %s", charm, channel_from_env)
```
sinapah: Very very minor: I wonder if we need to validate the channel env var? We can safely assume that the user enters a valid channel 99% of the time. Just something we can think about, but I doubt we actually need it.

tests/integration/test_log_rotation.py
```diff
@@ -9,9 +9,9 @@
 from constants import INTERNAL_TELEMETRY_LOG_FILE
 
 
-async def test_deploy(juju: jubilant.Juju, charm: str):
+def test_deploy(juju: jubilant.Juju, charm_22_04: str):
```
sinapah: QQ: this test don't use Zookeeper, so could we use base 24.04 for both the principal and the subordinate?

sinapah (approved):
Please also do a `fetch-lib`. Loki, Grafana, Tempo, Mimir!

PR #242, opened 2026-04-07
fix: Escape `$` in Prometheus scrape configs

MichaelThamm:
## Issue
Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/238

## Solution

- **`src/config_builder.py`**: Added `_sanitize_escape_prometheus_scrape_configs` static method to `ConfigBuilder` that recursively deep-copies and escapes bare `$` → `$$` (already-escaped `$$` is left unchanged). Called at pre-build (`build()`) time using `_sanitize_prometheus_scrape_configs` so all prometheus receivers have sanitized scrape jobs before writing to disk.

  Key regex: `re.sub(r'(?<!\$)\$(?!\$)', '$$', value)` — matches `$` not preceded or followed by another `$`, making the operation idempotent.

  ```python
  # Before: otelcol chokes on this
  {"replacement": "${1}.juju-73d163-3-lxd-0"}

  # After: otelcol renders $$ as literal $
  {"replacement": "$${1}.juju-73d163-3-lxd-0"}
  ```

- **`tests/unit/test_config_builder.py`**: Added parametrized tests covering bare `${N}`, multiple groups, `$VAR`-style, already-escaped `$$` (idempotency), no-dollar, and empty-string cases, plus explicit no-mutation (deep copy) and idempotency tests.

> [!NOTE]
> This solution accounts for users supplying external configs (otelcol-integrator -> `self._configure_external_configs`); they can specify Prometheus scrape configs without manually escaping the $ symbols.

This will prevent otelcol config loading error - scoped to Prometheus receiver `scrape_configs`

### Checklist

## Context

otelcol's [env-var expansion](https://opentelemetry.io/docs/collector/configuration/#environment-variables) applies to the entire config at load time. `$$` is the otelcol escape for a literal `$`, so doubling bare dollar signs in scrape config strings is the correct approach without modifying the actual Prometheus semantics at scrape time.

## Testing Instructions

1. Run `pytest tests/unit/test_config_builder.py` — all tests should pass.
2. To reproduce the original failure: relate a charm that injects `replacement: ${1}.something` via `cos_agent` and verify the collector no longer fails to start.


tests/unit/test_config_manager.py
```diff
@@ -310,6 +310,126 @@ def test_add_external_configs_adds_components_to_requested_pipelines():
...
+        ("$MY_VAR", "$$MY_VAR"),
+        # Already escaped $$ — must be idempotent
+        ("$${1}.juju-73d163-3-lxd-0", "$${1}.juju-73d163-3-lxd-0"),
+        # No dollar signs — unchanged
+        ("static-replacement", "static-replacement"),
+        # Empty string
+        ("", ""),
+    ],
+)
+def test_sanitize_escape_prometheus_scrape_configs_parametrized(
```
MichaelThamm: Update this test to use the _replace_dollars method because it is scoped to that.

The next test can assert that the rest of the config is not updated when running `sanitize_escape_prometheus_scrape_configs`.

Or we can combine both into one; likely the best

tests/unit/test_config_manager.py
```diff
@@ -310,6 +310,126 @@ def test_add_external_configs_adds_components_to_requested_pipelines():
...
+    ]
+
+    # WHEN sanitize_escape_prometheus_scrape_configs is called
+    result = ConfigManager.sanitize_escape_prometheus_scrape_configs(scrape_configs)
+
+    # THEN the replacement value has its bare $ escaped
+    assert result[0]["relabel_configs"][0]["replacement"] == "$${1}.juju-73d163-3-lxd-0"
+    # AND other fields are unchanged
+    assert result[0]["job_name"] == "juju_ceph-mon_prometheus_scrape"
+    assert result[0]["static_configs"] == [{"targets": ["10.0.0.1:9283"]}]
```
MichaelThamm:
```suggestion
    assert result == expected
```

tests/unit/test_config_manager.py
```diff
@@ -310,6 +310,126 @@ def test_add_external_configs_adds_components_to_requested_pipelines():
...
+
+
+def test_sanitize_escape_prometheus_scrape_configs_multiple_jobs_mixed():
```
MichaelThamm: This could be scoped to _escape_dollars

MichaelThamm (commented):
- [ ] Make sure to create a PR for track/2

src/config_builder.py
```diff
@@ -428,8 +431,46 @@ def _set_prometheus_receiver_global_timeout_and_interval(self, interval: str, ti
...
+    @staticmethod
+    def _escape_dollars(value: Union[str, list, dict]) -> Union[str, list, dict]:
+        """Recursively escape bare `$` signs in strings within a nested structure."""
+        if isinstance(value, str):
+            return re.sub(r'(?<!\$)\$(?!\$)', '$$', value)
+        if isinstance(value, dict):
+            return {k: ConfigBuilder._escape_dollars(v) for k, v in value.items()}
+        if isinstance(value, list):
+            return [ConfigBuilder._escape_dollars(item) for item in value]
+        return value
```
Abuelodelanada: What about using pattern matching instead on `isinstance` and `@classmethod` so we avoid calling `ConfigBuilder` directly?

```suggestion
    @classmethod
    def _escape_dollars(cls, value: Any) -> Any:
        """Recursively escape bare `$` signs in strings within a nested structure."""
        match value:
            case str():
                return re.sub(r'(?<!\$)\$(?!\$)', '$$', value)
            case dict():
                return {k: cls._escape_dollars(v) for k, v in value.items()}
            case list():
                return [cls._escape_dollars(item) for item in value]
            case _:
                return value
```

src/config_builder.py
```diff
@@ -428,8 +431,46 @@ def _set_prometheus_receiver_global_timeout_and_interval(self, interval: str, ti
...
+
+        Already-escaped `$$` sequences are left unchanged (idempotent).
+
+        Args:
+            scrape_configs: list of scrape config dicts (may contain nested dicts/lists/strings).
+
+        Returns:
+            A deep copy of the scrape configs with all bare `$` signs escaped to `$$`.
+        """
+        return [cast(Dict, ConfigBuilder._escape_dollars(copy.deepcopy(job))) for job in scrape_configs]
```
Abuelodelanada: As far as I understand `_escape_dollars` is returning a new structure, do we need `deepcopy`?

```suggestion
        return [cls._escape_dollars(job) for job in scrape_configs]
```

Abuelodelanada (changes requested):

src/config_builder.py
MichaelThamm: We want to ensure that the original job is not mutated. It is fine that the _escape_dollars method mutates but not the _sanitize_escape_prometheus_scrape_configs

Abuelodelanada (approved):

PR #243, opened 2026-04-07
refactor(track/2): Remove asyncio in itests

MichaelThamm:
## Issue

Like this PR:
- https://github.com/canonical/opentelemetry-collector-operator/pull/237

but for track/2

---

We do not need asyncio in our itests since jubilant is synchronous.

## Solution

Remove asyncio and use pytest-jubilant instead of pytest-operator [like we do in our monorepos](https://github.com/canonical/loki-operators/blob/f47f3595f98a92afd166690b05fc2f85e252b8a0/tests/integration/conftest.py#L28).

### Checklist

## Testing Instructions

See CI


sinapah (approved):
LGTM! Do we need a fetch-lib as well?

PR #247, opened 2026-04-13
fix: No metric normalization in prometheus components

MichaelThamm:
## Issue

Resolves partially https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/232

The core of the conversation will be held in:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/238

## Upgrade Notes

Any queries|dashboards that were created prior to adopting the otelcol charm and expecting the e.g., `_total` suffix (among other suffixes) need to be updated because we have disabled this in the `send-remote-write` integration.

### DIFF
--- a/src/config_manager.py
+++ b/src/config_manager.py
@@ -366,7 +366,10 @@ class ConfigManager:
         self.config.add_component(
             Component.receiver,
             f"prometheus/metrics-endpoint/{self._unit_name}",
-            config={"config": {"scrape_configs": jobs}},
+            config={
+                "config": {"scrape_configs": jobs},
+                "trim_metric_suffixes": True,
+            },
             pipelines=[f"metrics/{self._unit_name}"],
         )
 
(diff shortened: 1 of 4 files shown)


src/config_manager.py
```diff
@@ -380,6 +383,7 @@ def add_remote_write(self, endpoints: List[Dict[str, str]]):
                 {
                     "endpoint": endpoint["url"],
                     "tls": {"insecure_skip_verify": self._insecure_skip_verify},
+                    "add_metric_suffixes": False,
```
nobuto-m, line 383: Is this going to be hard-coded? Is there any way to opt-in to the new scheme once a principal charm has a proper suffix everywhere including the Grafana dashboard definitions?

Having the backward compatibility with grafana-agent by default is a good thing, but for the future, it might be a good idea to let a principal charm control this knob to declare the readiness for the new otel-collector scheme when they are ready to drop the support for grafana-agent at some point?

MichaelThamm: We have featured the send-otlp and receive-otlp relations in the collector charms which avoids the problematic prometheusremotewriteexporter entirely. We want to hardcode this to maintain parity with Grafana-agent to avoid breaking teams who migrated.

We could make this configurable, but we think this will add a maintenance burdon and users who want to use the new telemetry semantics of upstream Otel need to migrate to OTLP receivers and exporters. This will be featured for all COS charms (this is already featured for Otelcol charms) next cycle.

nobuto-m: > We have featured the send-otlp and receive-otlp relations in the collector charms which avoids the problematic prometheusremotewriteexporter entirely.

Ah, the future is switching the relation from `send-remote-write` -> `send-otlp`. That makes sense. Thanks for the clarification.

src/config_manager.py
```diff
@@ -380,6 +380,7 @@ def add_remote_write(self, endpoints: List[Dict[str, str]]):
...
```
Abuelodelanada, line 383: Even though we are doing this to keep parity with grafana-agent and to avoid breaking charm's dashboards, shouldn't we add a `warning` log message if this `add_metric_suffixes` is `False` saying something like: "`Counter` metrics without a `_total` suffix does not comply with otel standards. In the next otelcol charm major version `_total` suffix will be added. Please update your dashboards"?

I mean, at some point we want to follow otels standars...

Abuelodelanada (approved):

src/config_manager.py
MichaelThamm: But this is hardcoded, meaning that people even with valid metrics will receive this warning. Maybe there is a prometheus remote write exporter metric that would indicate when metrics suffixes were added, or using otel weaver, but this would need to be a spike.

Issue #129, opened 2025-11-18
otelcol runs havoc on CPU


erik78se:
### Bug Description

We recently had an issue with Loki, not being able to somehow process logs from the otelcol. The disk ran full on Loki side.

Somehow, this caused otelcol to run mad on CPU which is a very scary situation as if this happens in a production environment - its pretty much over. **DOWNTIME**

<img width="595" height="149" alt="Image" src="https://github.com/user-attachments/assets/69718826-40f9-4ac7-9dfb-b4a5f2c95d94" />

Looking at this situation, I wonder what we can do to safeguard from this to happen? As it is now, I just got afraid of deploying this into a large installation.

What can I do to help here? We're on LXD.

As an observation, the otelcollector is also restarting alot.... Possibly due the the error of "Sending queue is full?"

Nov 18 09:32:43 juju-b3adbc-2 opentelemetry-collector.opentelemetry-collector[180573]: 2025-11-18T09:32:43.727Z        error        adapter/receiver.go:61        ConsumeLogs() failed        {"resource": {"service.instance.id": "635ac96a-d500-465f-9279-c79fb67373d9", "service.name": "otelcol", "service.version": "0.130.0"}, "otelcol.component.id": "filelog/var-log", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "error": "sending queue is full"}

### To Reproduce

juju deploy opentelemetry-collector ....

### Environment

lxd

### Relevant log output

```shell
journalctl -xef -u snap.opentelemetry-collector.opentelemetry-collector.service 

Nov 18 09:27:43 juju-b3adbc-2 opentelemetry-collector.opentelemetry-collector[137598]:         github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza@v0.130.0/fileconsumer/internal/reader/reader.go:265
Nov 18 09:27:43 juju-b3adbc-2 opentelemetry-collector.opentelemetry-collector[137598]: github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/fileconsumer/internal/reader.(*Reader).ReadToEnd
Nov 18 09:27:43 juju-b3adbc-2 opentelemetry-collector.opentelemetry-collector[137598]:         github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza@v0.130.0/fileconsumer/internal/reader/reader.go:120
[...]

PR #251, opened 2026-04-16
feat: Memory limiting processor

MichaelThamm:
## Issue

Fixes partially https://github.com/canonical/opentelemetry-collector-operator/issues/129

I created this issue to address some knowledge gap:
- https://github.com/canonical/opentelemetry-collector-operator/issues/265

## Solution

From the upstream otelcol [when to scale docs](https://opentelemetry.io/docs/collector/scaling/#when-to-scale), it recommends using the [memorylimiterprocessor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor) “to prevent out of memory situations on the collector” by “performing periodic checks of memory usage and will begin refusing data and forcing garbage collection to reduce memory consumption when defined limits have been exceeded”.

Alternatives were considered like using [snap resource quotas](https://snapcraft.io/docs/how-to-guides/manage-snaps/use-resource-quotas/) (experimental feature) and k8s resource limits, but this would create a larger maintenance effort and less parity between the two charms. Ultimately, the processor is not as accurate regarding the hard limit like the snap and k8s limits, but this is a good tradeoff.

### Checklist
	- https://github.com/canonical/observability-stack/issues/268
- [x] We should update the doc to include the ability to override the mem limiting processor config using the `processors` config option.
	- https://github.com/canonical/observability-stack/pull/294
- [x] The user's processor config for mem limit should override ours ONLY FOR their specified pipelines.
- [x] Add utest for BlockedStatus when clamping
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/257

## Context

From the [best practices](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor#best-practices), “[this processor] is not a replacement for properly sizing and configuring the collector”. E.g., a memory limit of 1 MiB while running OpenStack, you will hit the memory limit and not OOM kill your host where the collector exists, but you have undersized the collector.

These will need to be addressed in the future:
- https://github.com/canonical/opentelemetry-collector-operator/issues/256
- https://github.com/canonical/opentelemetry-collector-operator/issues/260

References:
- https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html
- https://www.man7.org/linux/man-pages/man3/sysconf.3.html

### Get memory limits with python
I tested this in both VM and K8s charms:
```
ubuntu@juju-4bcff3-0:~$ python3
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> total_mib = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES") // (1024 * 1024)
>>> total_mib
88128
>>> 
ubuntu@juju-4bcff3-0:~$ free -m
               total        used        free      shared  buff/cache   available
Mem:           88128         143       86249           0        1735       87984
Swap:           8191           0        8191
```
### Manual charm testing

```bash
jd ubuntu --channel latest/stable --base ubuntu@24.04 ub
jd ./opentelemetry-collector_ubuntu@24.04-amd64.charm --trust otelcol
jrel otelcol ub
```
with no errors in the logs:
```
jssh otelcol/0 "tail -f /var/snap/opentelemetry-collector/common/otelcol.log"
jssh otelcol/0 "sudo snap logs -f opentelemetry-collector"
```
```bash
jssh otelcol/0 "free -m" 
               total        used        free      shared  buff/cache   available
Mem:           15986        4358        3626          10        8349       11627
Swap:              0           0           0

jssh otelcol/0 "cat /sys/fs/cgroup/memory.max"
max

jssh otelcol/0 "cat /etc/otelcol/config.d/otelcol_0.yaml" | yq -r '.processors.memory_limiter'
```
```yaml
{
  "memory_limiter": {
    "check_interval": "1s",
    "limit_mib": 15986,
    "spike_limit_mib": 1598
  },
}
```
check the Go runtime `MemStats.Alloc` (current heap in use) with Grafana:
image

This is averaging around `~3MiB`.

Now let's set the mem limit of the LXD VM to simulate a low-mem environment:
```bash
lxc config set juju-717e64-0 limits.memory 500MiB
jssh otelcol/0 "free -m"
               total        used        free      shared  buff/cache   available
Mem:             500         152          34           0         313         347
Swap:              0           0           0

jssh otelcol/0 "cat /sys/fs/cgroup/memory.max"
524288000

jc otelcol memory_limit_percentage=2
jssh otelcol/0 "cat /etc/otelcol/config.d/otelcol_0.yaml" | yq -r '.processors.memory_limiter'
```
```yaml
{
  "memory_limiter": {
    "check_interval": "1s",
    "limit_mib": 10,
    "spike_limit_mib": 2
  },
}
```

Check for logs indicating that the memorylimiterprocessor is in effect:
1. Soft limit reached:
> 2026-04-22T12:14:33.517Z        warn    memorylimiter@v0.130.1/memorylimiter.go:213     Memory usage is above soft limit. Refusing data.        {"resource": {"service.instance.id": "05b30820-2670-4710-b0db-a9fa010f1c6b", "service.name": "otelcol", "service.version": "0.130.1"}, "otelcol.component.id": "memory_limiter", "otelcol.component.kind": "processor", "otelcol.pipeline.id": "metrics/otelcol/0", "otelcol.signal": "metrics", "cur_mem_mib": 12}
2. Hard limit reached:
> 2026-04-22T12:08:55.177Z        warn    memorylimiter@v0.130.1/memorylimiter.go:196     Memory usage is above hard limit. Forcing a GC. {"resource": {"service.instance.id": "9f774ce2-bbdd-44b8-95aa-abe1bd6b72eb", "service.name": "otelcol", "service.version": "0.130.1"}, "otelcol.component.id": "memory_limiter", "otelcol.component.kind": "processor", "otelcol.pipeline.id": "traces/otelcol/0", "otelcol.signal": "traces", "cur_mem_mib": 12}
3. Garbage collcetion:
> 2026-04-22T12:00:54.179Z        info    memorylimiter@v0.130.1/memorylimiter.go:171     Memory usage after GC.  {"resource": {"service.instance.id": "9f774ce2-bbdd-44b8-95aa-abe1bd6b72eb", "service.name": "otelcol", "service.version": "0.130.1"}, "otelcol.component.id": "memory_limiter", "otelcol.component.kind": "processor", "otelcol.pipeline.id": "traces/otelcol/0", "otelcol.signal": "traces", "cur_mem_mib": 11}
4. Upstream pipeline component refuses telemetry:
> 2026-04-22T12:08:54.309Z        error   adapter/receiver.go:61  ConsumeLogs() failed    {"resource": {"service.instance.id": "9f774ce2-bbdd-44b8-95aa-abe1bd6b72eb", "service.name": "otelcol", "service.version": "0.130.1"}, "otelcol.component.id": "filelog/var-log", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "error": "data refused due to high memory usage"}

Notice that the memory usage of the collector drops:
image

## Upgrade Notes

New versions of this charm will default the hard limit to `100%` and the soft limit to `80%`. This means that collectors which are currently overloaded and operating at `> 80%` of total memory will start to reject incoming telemetry until they reduce the incoming load per collector.


src/charm.py
```diff
@@ -252,11 +256,16 @@ def _reconcile(self):
...
+        limits_processor = self._configure_limits_processor(config_manager)
+
         # Self-mon logging
         self._configure_logrotate()
 
         # Tracing setup
-        requested_tracing_protocols = integrations.receive_traces(self, tls=is_tls_ready(), ports=port_map)
+        requested_tracing_protocols = integrations.receive_traces(
+            self, tls=is_tls_ready(), ports=port_map
+        )
```
sed-i: On one hand, your formatting config breaks the line here, ...

src/charm.py
```diff
@@ -302,9 +311,7 @@ def _reconcile(self):
                             "scrape_interval": "60s",
                             "static_configs": [
                                 {
-                                    "targets": [
-                                        f"0.0.0.0:{port_map[Port.node_exporter.name]}"
-                                    ],
+                                    "targets": [f"0.0.0.0:{port_map[Port.node_exporter.name]}"],
```
sed-i: ...and on the other hand, it removes line breaks here.

In both cases, it's just noise for the PR.
We [already have an `fmt`](https://github.com/canonical/opentelemetry-collector-operator/blob/0131dc9e63aea068388c926eec8be4c6fbe0505b/tox.ini#L46) so if it's your IDE re-formatting the code, then perhaps it should be disabled? Unless we have this formatting back-and-forth because of changes in ruff itself?

src/charm.py
```diff
@@ -779,6 +792,22 @@ def _write_secrets_to_disk(self, external_secret_files: dict[str, str]) -> None:
...
+            if limit < 0 or limit > 100:
+                raise ValueError("memory_limit_percentage value must be [0, 100]")
+            config_manager.add_memory_limiter_processing(limit)
+        except ValueError:
+            logger.warning(
+                "Invalid memory_limit_percentage config value, defaulting to 100. "
+                "Valid values are [0, 100]"
+            )
+            config_manager.add_memory_limiter_processing(100)
+            return InvalidMemoryLimiterConfig()
```
sed-i: Seems odd here to:
- Use exception and immediately handle it as an alternative flow control.
- Return `None | InvalidMemoryLimiterConfig`, where `None` means success.
- Check for error with isinstance, instead of try-except.

I think we're still a bit away from `@final @dataclass(frozen=True) class Err(Result[E, A])`, if that's what you were aiming for.

src/constants.py
```diff
@@ -34,3 +34,4 @@
     "processes",
     "sysctl",
 }
+CGROUP_MEMORY_MAX: Final[str] = "/sys/fs/cgroup/memory.max"
```
sed-i: Iiuc, the path is missing the cgroup name, i.e.
- `/sys/fs/cgroup/memory.max` never exists;
- but `/sys/fs/cgroup/system.slice/memory.max` does exist.

I think snaps run under the `system.slice` or `user.slice` cgroup.

I wonder if quota groups are applied, then the `memory.max` value of the general `user.slice` cgroup is superseded.

Wdyt about obtaining it from the snap context? E.g.

```
$ snap run --shell opentelemetry-collector.opentelemetry-collector -c 'grep MemTotal /proc/meminfo'
MemTotal:       16356768 kB
```

I wonder if `snap run --shell -c` is available from the snap module. In any case we could raise a feature request for it (not blocking this work).

src/charm.py
```diff
@@ -779,6 +792,22 @@ def _write_secrets_to_disk(self, external_secret_files: dict[str, str]) -> None:
...
         config_manager.add_external_configs(self.external_configs)
 
+    def _configure_limits_processor(
+        self, config_manager: ConfigManager
+    ) -> None | InvalidMemoryLimiterConfig:
+        try:
+            limit = int(cast(str, self.config.get("memory_limit_percentage")))
+            if limit < 0 or limit > 100:
+                raise ValueError("memory_limit_percentage value must be [0, 100]")
+            config_manager.add_memory_limiter_processing(limit)
```
sed-i: Should we further split this out to something like
```python
limit = parse_memory_limit(self.config.get("..."))
config_manager.add_memory_limiter_processing(limit)
```
Seems like we'd gain in testability.

src/config_manager.py
```diff
@@ -1,18 +1,38 @@
...
 from config_builder import Component, ConfigBuilder, Port, build_port_map
-from constants import FILE_STORAGE_DIRECTORY
+from constants import CGROUP_MEMORY_MAX, FILE_STORAGE_DIRECTORY
 from integrations import ProfilingEndpoint
-from charmlibs.interfaces.otlp import OtlpEndpoint
 
 logger = logging.getLogger(__name__)
 
 
+def _total_memory_mib() -> int:
```
sed-i: Should we move this function to utils.py?

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+        data by returning errors to the preceding component in the pipeline.
+        When memory usage exceeds the hard limit, the processor additionally
+        forces garbage collection.
+
+        | user input (%) | hard limit (% of total) | soft limit (% of hard) |
+        | -10            | 0                       | 0                      |
+        | 50             | 50                      | 40                     |
+        | 100            | 100                     | 80                     |
+
+        Args:
```
sed-i, line 752: Add ref to upstream docs?

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+    def add_memory_limiter_processing(self, limit_percentage_request) -> None:
+        """Configure the memory limiter processor.
+
+        The time between measurements of memory usage is hardcoded to 1 second
+        as recommended by the processor's documentation.
+
+        The resulting limit_percentage (hard limit) is clamped to [0, 100] as a
+        percentage of total memory. The spike_limit_percentage is hardcoded to
+        20% of the hard limit. A hard limit of 0% means no memory limit is
+        enforced.
```
sed-i: We should rephrase this paragraph so it's clear that the 20% means that we get to keep 80%, and be more explicit about values, for example:

```diff
-A hard limit of 0% means
+when limit_percentage_request is 0
```

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
                     logger.debug(
                         "component type: '%s', name: '%s' added to config", config_type, comp_name
                     )
+
+    def add_memory_limiter_processing(self, limit_percentage_request) -> None:
+        """Configure the memory limiter processor.
+
+        The time between measurements of memory usage is hardcoded to 1 second
+        as recommended by the processor's documentation.
```
sed-i: This should be moved lower down, because when the IDE gives an autocomplete tooltip, it's most useful to get info on the public iface.

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+
+        The resulting limit_percentage (hard limit) is clamped to [0, 100] as a
+        percentage of total memory. The spike_limit_percentage is hardcoded to
+        20% of the hard limit. A hard limit of 0% means no memory limit is
+        enforced.
+
+        When memory usage exceeds the soft limit, the processor starts refusing
+        data by returning errors to the preceding component in the pipeline.
+        When memory usage exceeds the hard limit, the processor additionally
+        forces garbage collection.
```
sed-i: This should be more direct/explicit - "when soft limit" is too generalized, because this function hardcodes 20%, so we should probably say something like "when exceeding 80% of limit_percentage_request".

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+
+        | user input (%) | hard limit (% of total) | soft limit (% of hard) |
+        | -10            | 0                       | 0                      |
+        | 50             | 50                      | 40                     |
+        | 100            | 100                     | 80                     |
```
sed-i: It would be great to have this table as a cucumber test. Here it is hard to grok.

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+
+        Args:
+            limit_percentage_request: The hard limit, clamped to [0, 100],
```
sed-i: "clamped to [0, 100]" is not enforceable at the call site, and subject to processing inside the function. Should we drop that phrase?

src/config_manager.py
```diff
@@ -729,3 +749,47 @@ def add_external_configs(self, external_configs: List[Dict[str, Any]]) -> None:
...
+            as a percentage of total memory at which the processor forces GC.
+        """
+        total_mib = _total_memory_mib()
+        hard_limit_percentage = max(0, min(limit_percentage_request, 100))
+        hard_limit_mib = hard_limit_percentage * total_mib // 100
+        spike_limit_mib = hard_limit_mib * 20 // 100
+        self.config.add_component(
```
sed-i, line 759: What happens when the user specifies memory limiter in as a config option to the [processor config](https://github.com/canonical/opentelemetry-collector-operator/blob/0131dc9e63aea068388c926eec8be4c6fbe0505b/charmcraft.yaml#L204)? I guess we should add ours only if they didn't provide theirs? Also, need to update the description for the config option.

tests/unit/test_config_manager.py
```diff
@@ -332,3 +333,31 @@ def test_add_external_configs_skips_malformed_entries(external_configs):
...
+    spike_limit_mib = memory_limiter_config["spike_limit_mib"]
+    assert limit_mib >= 0
+    assert limit_mib <= 1024
+    assert spike_limit_mib >= 0
+    assert limit_mib >= spike_limit_mib
+
+    # AND the hard limit matches the clamped percentage of total memory
+    assert limit_mib == clamped_percentage * 1024 // 100
+    # AND the spike limit (soft-to-hard gap) is 20% of the hard limit
+    assert spike_limit_mib == limit_mib * 20 // 100
```
sed-i: These assertions somewhat replicate the logic being tested.
It's totally fine in a test like this to assert against literals.

tests/unit/test_config_manager.py
```diff
@@ -332,3 +333,31 @@ def test_add_external_configs_skips_malformed_entries(external_configs):
...
+
+    # THEN the memory limiter processor is added to the config with valid limits
+    assert "memory_limiter" in config_manager.config._config["processors"]
+    memory_limiter_config = config_manager.config._config["processors"]["memory_limiter"]
+    limit_mib = memory_limiter_config["limit_mib"]
+    spike_limit_mib = memory_limiter_config["spike_limit_mib"]
+    assert limit_mib >= 0
+    assert limit_mib <= 1024
+    assert spike_limit_mib >= 0
+    assert limit_mib >= spike_limit_mib
```
sed-i: These assertions seem redundant given the exact match below.

tests/unit/test_limits_processor.py
```diff
@@ -0,0 +1,87 @@
...
+    cfg = get_otelcol_config_file(unit_name, config_folder)
+    processor = cfg.get("processors", {}).get("memory_limiter", {})
+
+    # THEN the default hard limit (100%) is used and no error is raised
+    # * the spike limit is 20% of the hard limit
+    assert processor["limit_mib"] == 1024
+    assert processor["spike_limit_mib"] == 1024 * 20 // 100
+
+    # AND the charm is in BlockedStatus due to invalid config value
+    assert isinstance(state_out.unit_status, BlockedStatus)
```
sed-i: Should we have here a strict match?

```suggestion
    assert type(state_out.unit_status) is BlockedStatus
```

src/config_manager.py
MichaelThamm: This is a good point, currently we would accept both because theirs would be added under the `foo/custom/bar` namespace. However, I am not sure what happens if we have 2 processors defined for 1 pipeline. I can test this, or we can allow the user to override ours with theirs.

We should add a utest for this.

src/config_manager.py
MichaelThamm: I think its fine here because we are inside the function. We _can_ remove this from the `charmcraft.yaml`, although we have a feature that this is the case, so why not document it?

src/constants.py
MichaelThamm: In my testing, it does exist:
```
lxc config set juju-717e64-0 limits.memory 2GiB
jssh otelcol/0 "cat /sys/fs/cgroup/memory.max"
2147483648

jssh otelcol/0 "cat /sys/fs/cgroup/system.slice/memory.max"
max

jssh otelcol/0 "cat /sys/fs/cgroup/user.slice/memory.max"
max
```

We know from the failure logs that the collector tries to:
> open /sys/fs/cgroup/memory.max: permission denied

MichaelThamm:
For this PR:
- Replace the `os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES") // (1024 * 1024)` logic with something more solid e.g., it doesn't reflect the `lxc config` constraints
- Update the existing issue and ask in upstream otelcol how to obtain the mem for Linux systems and snaps

src/config_manager.py
```diff
@@ -633,6 +614,8 @@ def add_custom_processors(self, processors_raw: str) -> None:
                     f"traces/{self._unit_name}",
                 ],
             )
+            if processor_name == "memory_limiter":
+                self.config.remove_component("memory_limiter", Component.processor)
```
sed-i: Would it make sense to keep processors in a var,
```python
processors = yaml.safe_load(processors_raw)
```

so we could do a single member check instead?

```python
if processor_name in processors:
```

But I wonder if we have a potential code ordering problem here:
- Are we assuming that `add_custom_processors` runs after we process the config option?
- If `memory_limiter` is the one and only possible key (i.e. no `memory_limiter/some_label` possible), should we simply overwrite the value?
- The call here for `remove_component` happens after we add the custom processor. So would it remove both? The first one? The last?

tests/unit/test_config_builder.py
```diff
@@ -66,6 +65,63 @@ def test_add_to_pipeline(pipelines, component):
...
+    """remove_component removes the component definition and all pipeline references."""
+    # GIVEN a config with a component added to multiple pipelines
+    config = ConfigBuilder("", "", "", "")
+    pipelines = ["logs", "metrics", "traces"]
+    config.add_component(
+        component=component,
+        name="foo",
+        config={"a": "b"},
+        pipelines=pipelines,
+    )
```
sed-i: Is adding the same component twice of any interest for the test?

src/utils.py
```diff
@@ -1,4 +1,45 @@
-# Copyright 2025 Canonical Ltd.
+# Copyright 2026 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 """Utils module."""
+
+import logging
+import os
+
+from constants import CGROUP_MEMORY_MAX
```
sed-i: Nit: Should we move it out from constants into a function here,

```python
def _get_cgroup_memory_max_path():
```

with a comment that this should probably take a snap name or pid as arg?

src/utils.py
```diff
@@ -1,4 +1,45 @@
...
+
+    Reads the cgroup memory limit, which reflects container/LXD limits.
+    Falls back to physical RAM when the cgroup file is absent or reads "max"
+    (meaning no limit is set).
+
+    We need this function until this issue is closed:
+        - https://github.com/canonical/opentelemetry-collector-operator/issues/256
+    """
+    try:
+        raw = open(CGROUP_MEMORY_MAX).read().strip()
```
sed-i: Iiuc, we are missing a "close()" or a `with` here.
But you can accomplish very similar using [`pathlib.Path.read_text`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text).

src/utils.py
```diff
@@ -1,4 +1,45 @@
...
+def parse_memory_limit(config_value: str) -> int:
+    """Parse the memory limit percentage from config and validate it."""
+    try:
+        limit = int(config_value)
+        if limit < 0 or limit > 100:
+            raise ValueError("memory_limit_percentage value must be [0, 100]")
+        return limit
+    except ValueError:
+        logger.warning(
+            "Invalid memory_limit_percentage config value, defaulting to 100. "
```
sed-i: In the scope of this function, if we're "defaulting to 100", then why are we raising?

src/utils.py
```diff
@@ -1,4 +1,45 @@
...
+            "Valid values are [0, 100]"
+        )
+        raise
```
sed-i: Can we simplify this structure? E.g.
```python
try:
    limit = int(config_value)
except (ValueError, TypeError):
    logger.warning("...")
    raise

if limit < 0 or limit > 100:
    logger.warning("...")
    raise ValueError("...")

return limit
```

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,59 @@
+Feature: Memory limiter processor configuration
+  The opentelemetry-collector charm configures a memory limiter processor
+  to prevent out-of-memory scenarios. The user sets a hard limit as a
+  percentage of total memory via the memory_limit_percentage Juju config
+  option. The soft limit is always 80% of the hard limit (spike = 20%).
+
+  Background:
+    Given total available memory is 1024 MiB
```
sed-i: Can we also parse out 1024 from here?

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,59 @@
...
+
+  # --- Config manager unit tests ---
+
+  Scenario Outline: Hard limit is clamped to [0, 100] and spike is 20% of hard limit
+    Given a user requested hard limit percentage of <user_input>
+    When the memory limiter processor is added to the config
+    Then the hard limit is <clamped_percentage>% of total memory
+    And the spike limit is 20% of the hard limit
```
sed-i: Parse the `20`?

charmcraft.yaml
```diff
@@ -367,6 +367,36 @@ config:
         Example — override only the ports that conflict with other services:
           juju config opentelemetry-collector ports="loki_http=3501,otlp_grpc=4320"
       default: ""
+    memory_limit_percentage:
+      type: string
```
sed-i:
```suggestion
      type: int
```

charmcraft.yaml
```diff
@@ -367,6 +367,36 @@ config:
...
+      description: |
+        Request a hard limit for the collector's memory limiter processor as a
+        percentage of the total available memory. The resulting processor
+        config will be applied to all of the collector's pipelines.
+        
+        Ref: https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor
+        
+        The collector measures memory usage every second to determine if the
+        limits are being exceeded. The calculated (by the charm) hard limit is
+        clamped to [0, 100] percent of the total available memory and the soft
```
sed-i: Are we clamping or blocking?

tests/unit/test_memory_limiter_bdd.py
sed-i, line 1: - Take a look at [`target_fixture`](https://pytest-bdd.readthedocs.io/en/stable/). It could simplify a couple places and avoid relying on the `context` dict. LLMs are stuck on this `context` dict for some reason.
- It is not necessary to have a `given` step, you can jump straight to `when`.

charmcraft.yaml
MichaelThamm: Wow, I was convinced there was no `int` support in charmcraft.yaml TIL.

src/charm.py
```diff
@@ -252,11 +253,16 @@ def _reconcile(self):
             ports=port_map,
         )
 
+        # Memory limiter setup
+        valid_memory_limit = self._configure_limits_processor(config_manager)
```
sed-i: Can we bring this closer to the status setting?
I.e. why not move this down or bring the other one up?
If this is because the limits processor needs to be the first one conf'd, then we should update the logic so that order doesn't matter.

src/utils.py
```diff
@@ -1,4 +1,44 @@
...
+    try:
+        raw = _get_cgroup_memory_max_path().read_text().strip()
+        if raw != "max":
+            return int(raw) // (1024 * 1024)
+    except (OSError, ValueError):
+        pass
+    return os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES") // (1024 * 1024)
+
+
+def parse_memory_limit(config_value: int) -> int:
```
sed-i: There's no parsing and no clamping taking place here. If that's the intention then should this be called `validate_...`?

src/charm.py
```diff
@@ -781,6 +791,22 @@ def _write_secrets_to_disk(self, external_secret_files: dict[str, str]) -> None:
...
+        """
+        try:
+            limit = parse_memory_limit(cast(int, self.config.get("memory_limit_percentage")))
+        except ValueError:
+            limit = 100
+        else:
+            config_manager.add_memory_limiter_processor(limit)
+            return True
+        config_manager.add_memory_limiter_processor(limit)
+        return False
```
sed-i: Would this be tidier if we split things out?

```python
raw_limit = self.config.get("...")
limit = self._clamp_memory_limit(raw_limit)  # logs warning and clamps
config_manager.add_memory_limiter_processor(limit)
valid = raw_limit == limit
```

src/config_manager.py
```diff
@@ -603,6 +604,13 @@ def add_custom_processors(self, processors_raw: str) -> None:
         the OpenTelemetry Collector configuration.
         """
         for processor_name, processor_config in yaml.safe_load(processors_raw).items():
+            # "memory_limiter" is the name (and type) of the default processor we add
+            if processor_name == "memory_limiter" or processor_name.startswith("memory_limiter/"):
+                self.config.remove_component("memory_limiter", Component.processor)
+                logger.info(
```
sed-i: Debug?

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
+Feature: Memory limiter processor configuration
+  The opentelemetry-collector charm configures a memory limiter processor
+  to prevent out-of-memory scenarios. The user sets a hard limit as a
+  percentage of total memory via the memory_limit_percentage Juju config
+  option. The soft limit is always 80% of the hard limit (spike = 20%).
+
+  Background:
+    Given total available memory is 1024 MiB
```
sed-i: Now the 1024 seems unused. If you do use it, it probably should be parsed out.

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
...
+    And the spike percentage is 20
```
sed-i: Parse this out?

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
...
+
+  # --- Config manager unit tests ---
+
+  Scenario Outline: Hard limit is clamped to [0, 100] and spike is a percentage of hard limit
+    Given a user requested hard limit percentage of <user_input>
+    When the memory limiter processor is added to the config
+    Then the hard limit is <clamped_percentage>% of total memory
+    And the spike limit is 20 percent of the hard limit
```
sed-i: Parse out the 20? (why is it repeated on top of the background?)

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
...
+      | -10        | 0                  |
+      | 0          | 0                  |
+      | 60         | 60                 |
+      | 100        | 100                |
+      | 110        | 100                |
+
+  # --- Charm-level integration tests ---
+
+  Scenario: Default config sets hard limit to 100%
+    Given a user provides no value for the memory_limit_percentage config option
```
sed-i, line 32: Would it be possible to make this parsable (reusable)?

```suggestion
    Given a user provides memory_limit_percentage config option "None"
```

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
...
+
+  Scenario Outline: Out-of-range config values cause BlockedStatus
+    Given a user provides <value> for the memory_limit_percentage config option
+    When any event executes the reconciler
+    Then the charm is in BlockedStatus
+
+    Examples:
+      | value |
+      | -10   |
+      | 110   |
```
sed-i: Should we merge this in the test above by adding another column "charm_status"?

tests/unit/features/memory_limiter.feature
```diff
@@ -0,0 +1,53 @@
...
+
+  Scenario: User-provided memory_limiter overrides the default
+    Given a user provides a memory_limiter in the processors config option
```
sed-i, line 43: Would it make sense to inline the custom config here as a docstring? `docstring` is a pytest-bdd fixture you can use in a test.

charmcraft.yaml
```diff
@@ -367,6 +367,37 @@ config:
...
+        Ref: https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor
+        
+        The collector measures memory usage every second to determine if the
+        limits are being exceeded. If the requested limit is invalid, the charm
+        goes into blocked status and the hard limit defaults to 100% of the
+        total available memory. The soft limit is always 80% of the hard limit
+        percentage. A requested limit of 0% means no memory limit is enforced.
+        For example:
+
+        | user input (%) | hard limit (% of total) | soft limit (% of hard) |
```
sed-i: > soft limit (% of hard)

this is a bit confusing, because the soft limit is always 20%, so "% of hard" should remain 20 throughout, but the table indicates % of total I guess?

tests/unit/features/memory_limiter.feature
MichaelThamm: It is? I am confused by what you mean "parse this out"? If you look at the statement:
```python
@given(parsers.parse("the spike percentage is {spike:d}"), target_fixture="spike_percentage")
```

tests/unit/features/memory_limiter.feature
MichaelThamm: It is?
```python
@then(parsers.parse("the spike limit is {percentage:d} percent of the hard limit"))
```

> why is it repeated on top of the background?

The one from the background:
> And the spike percentage is 20

is different from the one in the assertions:
> And the spike limit is 20 percent of the hard limit

Because the first is setting `target_fixture="spike_percentage"` and the second is asserting that the percentage is correct.

src/charm.py
```diff
@@ -791,21 +788,25 @@ def _write_secrets_to_disk(self, external_secret_files: dict[str, str]) -> None:
     def _configure_external_configs(self, config_manager: ConfigManager):
         config_manager.add_external_configs(self.external_configs)
 
+    def _clamp_memory_limit(self, config_value: int) -> int:
+        """Default the memory limit percentage to 100 if input is not in [0, 100]."""
+        if config_value < 0 or config_value > 100:
+            logger.warning(
+                "Invalid memory_limit_percentage config value. Valid values are [0, 100]"
```
sed-i: Nit: `Valid values are [0, 100]` may not come across as a range. Let's make it more explicit?

```diff
-Invalid memory_limit_percentage config value. Valid values are [0, 100]
+Invalid memory_limit_percentage charm config option, must be in the range [0, 100].
```

src/config_builder.py
```diff
@@ -296,6 +296,7 @@ def add_component(
         name: str,
         config: Dict[str, Any],
         pipelines: Optional[List[str]] = None,
+        first_in_pipeline: bool = False,
```
sed-i: Typically this is done with an int priority value.
But I wonder if we should at all leak this outside?
Shouldn't `self._add_to_pipeline()` always ensure that memory limiter is the first?

src/config_builder.py
MichaelThamm, line 1: We should add to the `build` method functionality that checks if ANY `memory_limiter` processor was configured. IFF not, then add default. This protects us from the 2 cases of mem limiters in config:
1. custom processors
2. otelcol-integrator config

This removes the need for defensive config_manager methods and we then have the entire pre-rendered config so we can determine where to place that memory_limiter e.g., pos 1.

sed-i (approved):

PR #272, opened 2026-05-07
feat: add charms.just blueprints for changelog

lucabello:
Add a `charms.just` central blueprint with a `changelog` recipe and import it here.


sinapah (approved):
Thanks @lucabello, this looks great!

A few thoughts from me:
1. Should we create a dir named `change logs` and place change logs in it with the format `track-x-changelog-md`?
2. Would it be a good idea to sort features and fixes from most to least recent? The PR numbers are next to each issue so I think it would look nice if the entries were ordered.

PR #302, opened 2026-06-04
fix: forward related-application alert rules over `send-otlp`

Abuelodelanada:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-operator/issues/297

In tandem with: https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/274

When opentelemetry-collector (machine) is related to a downstream collector over the `send-otlp`/`receive-otlp` interface, alert rules from **related applications** (e.g. `postgresql`, related over `cos-agent`/`metrics-endpoint`) were not present in the `send-otlp` relation data. Only the collector's own bundled rules were forwarded, so downstream Prometheus/Loki never received marker alerts such as `PostgresqlInvalidIndex`.

## Solution

`send_otlp` published only the charm's **bundled** rules: it built the OTLP `RuleStore` from the source rule directories (`*_RULES_SRC_PATH`). The alert rules collected from related applications are staged into the **destination** rule directories (`*_RULES_DEST_PATH`) by other integrations, so they were never picked up. On top of that, `send_otlp` ran *before* those integrations had populated the destination directories.

This PR:

- **Reads the staged (DEST) rule directories** in `send_otlp` instead of the bundled (SRC) ones.
  The DEST directories already contain the bundled rules **plus** the alerts gathered from related applications:
  - `scrape_metrics` → `metrics-endpoint` alerts (`prometheus_alert_rules/`)
  - `receive_loki_logs` → `receive-loki-logs` alerts (`loki_alert_rules/`)
  - the `cos-agent` handling in `_reconcile` (leader-only) → cos-agent metrics/logs alerts
- **Reorders `_reconcile`** so `send_otlp` runs *after* the logs/metrics/cos-agent/remote-write
  integrations, guaranteeing the DEST directories are fully populated before the OTLP databag is published.
- **Hardens `cleanup()`** to resolve the rule directories against the absolute charm directory
  (now takes a `charm_root: Path`), matching the paths used by every integration that reads/writes them and making it independent of the process' current working directory.

### Checklist

## Context

- The collector forwards alert rules by **reading rule files on disk**. Integrations copy bundled rules and write related-app rules into the `*_RULES_DEST_PATH` directories; the forwarding integrations (`send_remote_write`, `send_loki_logs`, and now `send_otlp`) read those same directories. The ordering between *staging* and *reading* is therefore load-bearing.
- Note on `forward_alert_rules`: the `metrics-endpoint` and `receive-loki-logs` staging is gated on this config, but the `cos-agent` alert staging in `_reconcile` is gated only on `is_leader()` (pre-existing behaviour, consistent with the cos-agent → `send-remote-write` path). As a result, cos-agent alerts are forwarded over OTLP regardless of `forward_alert_rules`. This is documented in the new tests and left unchanged here to keep the fix scoped to #297.

## Testing Instructions

Unit tests:

```bash
tox -e unit
# or, just the OTLP feature:

tox -e unit -- -k test_otlp.py
```

New/updated tests in `tests/unit/test_otlp.py`:
- `test_related_app_alerts_forwarded_over_otlp` — related-app alert (via `metrics-endpoint`) is in
  the `send-otlp` databag, and its `juju_application` topology label is preserved.
- `test_related_app_alerts_forwarded_over_otlp_via_cos_agent` — same, exercising the real
  `cos-agent` subordinate path used by postgresql in production.
- `test_related_app_log_alerts_forwarded_over_otlp` — log (logql) alerts received over
  `receive-loki-logs` are forwarded.
- `test_reconcile_stages_otlp_rules_after_related_app_staging` — guards the reconcile call order
  (`send_otlp` runs after `cleanup` and after every related-app staging step).

To confirm the tests catch the regression, revert only `src/` and keep the test changes:

```bash
git checkout main -- src/charm.py src/integrations.py

tox -e unit -- -k test_otlp.py
# Expected: the new related-app forwarding tests FAIL
```

End-to-end (reproduces the issue scenario):

1. K8s model: deploy `opentelemetry-collector-k8s` + `prometheus`, offer the `receive-otlp`
2. Machine model: deploy `postgresql` + `opentelemetry-collector`, relate them over `cos-agent`. Consume the `receive-otlp` offer.
3. Verify the postgresql alert ruler reaches Prometheus:

## Upgrade Notes

No action required. After upgrading, alert rules from related applications (e.g. postgresql) will
start appearing in the `send-otlp` relation data on the next reconcile and propagate downstream. No
configuration or relation changes are needed.

### DIFF
--- a/lib/charms/tls_certificates_interface/v4/tls_certificates.py
+++ b/lib/charms/tls_certificates_interface/v4/tls_certificates.py
@@ -76,7 +76,7 @@ LIBAPI = 4
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 28
+LIBPATCH = 30
 
 PYDEPS = [
     "cryptography>=43.0.0",
@@ -1168,6 +1168,26 @@ class CertificateRequestAttributes:
             and self.additional_critical_extensions == other.additional_critical_extensions
         )
 
+    def __hash__(self) -> int:
+        """Return hash of the CertificateRequestAttributes object."""
+        return hash(
+            (
+                self.common_name,
+                frozenset(self.sans_dns) if self.sans_dns else None,
+                frozenset(self.sans_ip) if self.sans_ip else None,
+                frozenset(self.sans_oid) if self.sans_oid else None,
+                self.email_address,
+                self.organization,
+                self.organizational_unit,
+                self.country_name,
+                self.state_or_province_name,
+                self.locality_name,
+                self.is_ca,
+                self.add_unique_id_to_subject_name,
+                tuple(self.additional_critical_extensions),
+            )
+        )
+
     def is_valid(self) -> bool:
         """Validate the attributes of the certificate request.
 
@@ -1733,6 +1753,7 @@ class TLSCertificatesRequiresV4(Object):
                 "Invalid renewal relative time. Must be between 0.5 and 1.0"
             )
         self._private_key = private_key
+        self._cached_private_key: Optional[PrivateKey] = None
         self.renewal_relative_time = renewal_relative_time
         self.framework.observe(charm.on[relationship_name].relation_created, self._configure)
         self.framework.observe(charm.on[relationship_name].relation_changed, self._configure)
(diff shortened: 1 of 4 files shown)


MichaelThamm (commented):
This is great and I think this was the regression:
- https://github.com/canonical/opentelemetry-collector-operator/pull/234

I also think that this will be an issue in the K8s charm due to the same regression but in the K8s charm:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/222

if we implement `receive-otlp` in the future, we will have to remember this compatibility detail: writing the rules to disk.

MichaelThamm (approved):

PR #323, opened 2026-06-19
fix: restart otelcol snap when trusted CAs change on disk

Abuelodelanada:
## Issue

This PR fixes: https://github.com/canonical/opentelemetry-collector-operator/issues/304

A running otelcol snap caches its trusted CA store at startup. When a new CA is delivered via the `receive-ca-cert` relation and `update-ca-certificates` folds it into the system bundle, the snap does not restart, producing `x509: certificate signed by unknown authority` errors until a manual restart occurs.

The root cause is that the restart trigger was tied to the relation-derived hash returned by `integrations.receive_ca_cert()`. The `certificate_transfer` v1 handshake can split CA delivery across multiple `relation-changed` hooks, so the relation hash can settle on an earlier hook — one that runs *before* the hook that materialises the CA files and runs `update-ca-certificates`. By the time the trust store is actually updated, the trigger sees an unchanged hash and skips the restart.

## Solution

Change the restart trigger from hashing the *relation data* to hashing the CA files materialised on disk under `RECV_CA_CERT_FOLDER_PATH`. A new helper `hash_ca_cert_dir(folder)` (in `utils.py`) reads all `*.crt` files in the directory, sorted by name, and produces a stable SHA-256 of their names and contents. This hash only changes when the files on disk genuinely change, and it is computed *after* `receive_ca_cert` writes the certs and `refresh_certs` runs `update-ca-certificates` — keeping the restart aligned with the bytes the snap will actually load.

### Checklist

## Context

The `certificate_transfer` v1 interface (`receive-ca-cert`) uses a two-step handshake: the requirer announces `version: 1`, and the provider responds with the CA in its relation data. These can arrive in separate `relation-changed` events, meaning the CA is visible in the relation data on a hook that may not be the one where `update-ca-certificates` runs.

The `RECV_CA_CERT_FOLDER_PATH` is `/usr/local/share/ca-certificates/juju_receive-ca-cert`. Both the `receive-ca-cert` and `receive-server-cert` relations write their CA files into this directory, so the on-disk hash captures changes from either source.

## Testing Instructions

1. Run the unit test suite:
   ```shell
   tox run -e unit
   ```
   All tests should pass. The new tests in `tests/unit/test_ca_cert_restart.py` specifically verify:
   - `test_hash_*` — the hashing primitive behaves correctly.
   - `test_snap_restarted_when_ca_arrives` — a CA delivered over the relation restarts the snap.
   - `test_snap_restarted_when_on_disk_ca_changes_but_relation_hash_is_stale` — the race in #304 is fixed: the snap restarts even when the relation-derived hash does not change between hooks.
   - `test_snap_restarted_when_ca_changes_even_if_config_unchanged` — a changed CA restarts the snap even when config is stable.
   - `test_snap_not_restarted_when_nothing_changes` — no unnecessary restarts on no-op reconciles.
   - `test_snap_restarted_when_server_cert_changes` — a changed server cert also triggers a restart.

2. Deploy the charm with the `receive-ca-cert` relation and verify the otelcol snap restarts automatically when a new CA is provided (no manual restart needed).

## Upgrade Notes

No special upgrade steps required. The change is purely behavioural — the mechanism for deciding whether to restart the snap now examines on-disk CA files instead of relation data. Existing deployments will pick up the fix on their next reconcile (triggered by any hook).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -26,6 +26,7 @@ from tenacity import retry, stop_after_attempt, wait_fixed
 import integrations
 from config_builder import Component, Port, build_port_map
 from config_manager import ConfigManager
+from utils import hash_ca_cert_dir
 from constants import (
     CERT_DIR,
     CONFIG_FOLDER,
@@ -195,8 +196,11 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         if self.unit.is_leader():
             integrations.cleanup(self.charm_dir.absolute())
 
-        # Integrate with TLS relations
-        receive_ca_certs_hash = integrations.receive_ca_cert(
+        # Integrate with TLS relations.
+        # NOTE: receive_ca_cert writes the CA files under RECV_CA_CERT_FOLDER_PATH; its
+        # returned hash is intentionally not used as a restart trigger (see the
+        # hash_ca_cert_dir call below, which hashes the materialized files instead).
+        integrations.receive_ca_cert(
             self,
             recv_ca_cert_folder_path=LocalPath(RECV_CA_CERT_FOLDER_PATH),
         )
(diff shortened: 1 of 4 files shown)


sinapah (commented):
Looks good, should we add an integration test too?

lucabello (approved):
LGTM; this specific behavior seems not trivial to integration test, but feel free to add one as Sina mentioned!

PR #327, opened 2026-06-24
fix: apply path_exclude config to snap log jobs

lucabello:
Closes #324

The `path_exclude` configuration option was only applied to the `/var/log` scrape job, not to fstab-based snap log jobs. This caused high resource consumption when related snaps (like postgresql) rotate logs in the same folder.

It was agreed with @sed-i to only implement this for now, as it should address the problem with a smaller blast radius than modifying the cos agent interface when we're very close to release.


sed-i (approved):
LGTM!

PR #340, opened 2026-06-26
fix: use rate function instead of irate when using $__rate_interval

Abuelodelanada:
## Issue

Based on the finding made on [this comment of this issue](https://github.com/canonical/mimir-operators/issues/82#issuecomment-4804589909) we have discovered that according to [Grafana documentation](https://grafana.com/docs/grafana/latest/datasources/prometheus/template-variables/#use-__rate_interval) `$__rate_interval` was ment to be used with `rate` not `irate`:

> `$__rate_interval` is a Grafana-specific variable designed for use with `rate()` and `increase()`. It guarantees a range window large enough to capture at least four scrape samples, preventing gaps or inaccuracies in results.

## Solution

Change `irate` with `rate` where `$__rate_interval` is used

### Checklist

## Testing Instructions

1. Deploy cos-lite in a K8s model
2. Deploy otelcol in a VM model
3. Consume `grafana-dashboards` and `prometheus-receive-remote-write`
4. Relate otelcol to ubuntu charm
5. Relate otelcol to `grafana-dashboards` and `prometheus-receive-remote-write`
6. Log into Grafana
7. Verify that the modified Dashboards works well

### DIFF
--- a/src/grafana_dashboards/node-exporter-full.json
+++ b/src/grafana_dashboards/node-exporter-full.json
@@ -199,7 +199,7 @@
             "uid": "${prometheusds}"
           },
           "editorMode": "code",
-          "expr": "(sum by(instance) (irate(node_cpu_seconds_total{instance=\"$node\",job=\"$job\", mode!=\"idle\"}[$__rate_interval])) / on(instance) group_left sum by (instance)((irate(node_cpu_seconds_total{instance=\"$node\",job=\"$job\"}[$__rate_interval])))) * 100",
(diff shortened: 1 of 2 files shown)


sed-i (commented):
Per [docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#irate),

> irate should only be used when graphing volatile, fast-moving counters.

I imagine that is why the original authors of this dashboard chose irate.
And the dashboard is indeed working as expected in some contexts.

I think the problem is not irate itself, but again, the scrape interval.
If I were to rewrite their implementation in python code, I think it would be:

```python
def rate(v, t):
    dv = v[-1] - v[0]
    dt = t[-1] - t[0]
    return dv/dt

def irate(v, t):
    dv = v[-1] - v[-2]
    dt = t[-1] - t[-2]
    return dv/dt
```

but since `__rate_interval` is predefined, could there be some interference with the `dv/dt` calculation? Because `__rate_interval` can be less than, equal to, or greater than `dt`.

All that is to say that I don't understand why `irate(...)[$__rate_interval]` is different from `rate(...)[$__rate_interval]`.

PR #359, opened 2026-07-13
fix: COSagent dashboard title collision

MichaelThamm:
### Checklist
- [ ] I have added or updated relevant documentation.
- [ ] PR title makes an appropriate release note and follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) syntax.
- [ ] Merge target is the correct branch, and relevant tandem backport PRs opened.

### DIFF
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -409,7 +409,7 @@ def _get_dashboards(relations: List[Relation]) -> List[Dict[str, Any]]:
     return list(aggregate.values())
 
 
-def _add_dashboards(dashboards: List[Dict[str, str]], dest_path: Path):
+def _add_dashboards(dashboards: List[Dict[str, Any]], dest_path: Path):
     """Save the dashboards to files in the specified destination folder.
 
     For K8s charms, dashboards are saved in the charm container.
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -426,13 +426,22 @@ def _add_dashboards(dashboards: List[Dict[str, str]], dest_path: Path):
     """
     dest_path.mkdir(parents=True, exist_ok=True)
     for dash in dashboards:
-        # Build dashboard custom filename
+        # Build dashboard custom filename.
+        # The (title, charm, rel_id) triple is not guaranteed to be unique: multiple
+        # untitled dashboards, or dashboards sharing a title, from the same principal
+        # would otherwise collide on the same filename and silently overwrite each other
+        # (see https://github.com/canonical/cos-proxy-operator/pull/241). To make the
+        # filename collision-proof regardless of the title, we append a stable content
+        # identity: the dashboard `uid` if present, else a short hash of the content.
+        content = dash["content"]
         charm_name = dash.get("charm", "charm-name")
         rel_id = dash.get("relation_id", "rel_id")
         title = dash.get("title", "").replace(" ", "_").replace("/", "_").lower()
-        filename = f"juju_{title}-{charm_name}-{rel_id}.json"
+        uid = content.get("uid") or content.get("dashboard", {}).get("uid")
+        identity = uid or sha256(json.dumps(content, sort_keys=True))[:8]
+        filename = f"juju_{title}-{charm_name}-{rel_id}-{identity}.json"
         with open(Path(dest_path, filename), mode="w", encoding="utf-8") as f:
-            f.write(json.dumps(dash["content"]))
+            f.write(json.dumps(content))
             logger.debug("updated dashboard file %s", f.name)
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_cos_agent_dashboards.py
```diff
@@ -0,0 +1,161 @@
...
+            return json.loads(rel.local_app_data["dashboards"])["templates"]
+    raise AssertionError("no grafana-dashboards-provider relation found in output state")
+
+
+def test_multiple_no_title_dashboards_are_not_overwritten(ctx):
+    """Scenario: seven untitled cos-agent dashboards all survive to Grafana."""
+    # GIVEN seven dashboards from the same app, none of which have a title
+    # (the root scenario from cos-proxy#241: multiple untitled dashboards that
+    # previously all mapped to `juju_no_title-...json` and clobbered one another)
+    no_title_dashes = [{"uid": f"uid_{i}", "overwrite": True, "tags": []} for i in range(7)]
```
mmkay: While this is exactly the state from the cos-proxy issue, it's probably enough to have three here. It doesn't hurt to have more though :)

tests/unit/test_cos_agent_dashboards.py
```diff
@@ -0,0 +1,161 @@
...
+
+def _forwarded_dashboards(state_out) -> dict:
+    for rel in state_out.relations:
+        if rel.endpoint == "grafana-dashboards-provider":
+            return json.loads(rel.local_app_data["dashboards"])["templates"]
+    raise AssertionError("no grafana-dashboards-provider relation found in output state")
+
+
+def test_multiple_no_title_dashboards_are_not_overwritten(ctx):
+    """Scenario: seven untitled cos-agent dashboards all survive to Grafana."""
```
mmkay:
```suggestion
    """Scenario: seven untitled cos-agent dashboards all appear in Grafana."""
```

tests/unit/test_cos_agent_dashboards.py
```diff
@@ -0,0 +1,161 @@
...
+    # THEN the nested title is used instead of a `no_title` placeholder
+    templates = _forwarded_dashboards(state_out)
+    assert any("my_nested_dashboard" in k for k in templates), (
+        f"nested title not used for filename; got keys: {list(templates)}"
+    )
+    assert not any("no_title" in k for k in templates)
+
+
+def test_add_dashboards_same_title_do_not_overwrite(tmp_path):
+    """Defense in depth: `_add_dashboards` must not clobber same-title dashboards.
```
mmkay: As a non-native I struggle with understanding this part. Perhaps it would be enough to write something like "Test that dashboards with same title don't overwrite each other in _add_dashboards"?

src/integrations.py
```diff
@@ -426,13 +426,22 @@ def _add_dashboards(dashboards: List[Dict[str, str]], dest_path: Path):
...
+        # would otherwise collide on the same filename and silently overwrite each other
+        # (see https://github.com/canonical/cos-proxy-operator/pull/241). To make the
+        # filename collision-proof regardless of the title, we append a stable content
+        # identity: the dashboard `uid` if present, else a short hash of the content.
+        content = dash["content"]
         charm_name = dash.get("charm", "charm-name")
         rel_id = dash.get("relation_id", "rel_id")
         title = dash.get("title", "").replace(" ", "_").replace("/", "_").lower()
-        filename = f"juju_{title}-{charm_name}-{rel_id}.json"
+        uid = content.get("uid") or content.get("dashboard", {}).get("uid")
```
mmkay, line 440: Please have a look at the static analysis results. I think this will work but perhaps worth being a bit defensive in case the databag contents are surprising.

mmkay (commented):
A few comments as we're still in draft, curious what the tests will tell.

src/integrations.py
MichaelThamm: In cos-agent we have [this same logic](https://github.com/canonical/grafana-agent-operator/blob/4ad98c84e17ffbdeaea0851f9ed64f30f1003d27/lib/charms/grafana_agent/v0/cos_agent.py#L1407) (but for `uid` instead of `title`) on `content` which is the source of `dashboards`:
```python
title = content.get("title") or content.get("dashboard", {}).get("title")
```

MichaelThamm: Which is because of this typing in cos-agent:
- https://github.com/canonical/grafana-agent-operator/blob/4ad98c84e17ffbdeaea0851f9ed64f30f1003d27/lib/charms/grafana_agent/v0/cos_agent.py#L1391-L1396

I think the code is defensive and maybe we need to update the return type since we clearly don't know if the content can contain a nested `dashboard` key or not. I assume this was due to some schema change and for backwards compatibility?

> Grafana provisioning envelope format with the title inside the nested `dashboard` sub-object rather than at the top level

mmkay: I think this is fine in its current form :)

mmkay (approved):
Clever use of the hash!

PR #372, opened 2026-08-06
fix: re-create snap lockfiles on every reconcile

### LINKED ISSUE #208: LCK file isn't re-created by the charm if it was manually deleted
### Bug Description

Via @MoonC432
otelcol charm was refreshed, and ended up in blocked status with ["mismatched snap revision"](https://github.com/canonical/opentelemetry-collector-operator/blob/f62b0bb176c0ef521913c90a1cd06a8f8a9ada26/src/charm.py#L531).
As a troubleshooting attempt, node-exporter lock file was deleted. On config-changed, the node-exporter snap was installed, but no LCK file created.

### To Reproduce

SSH into the machine, delete the node-exporter lockfile in `/opt/singleton_snaps/...`, and run `juju config ..` to trigger an event.

### Environment

Latest track/2 as of time of writing.

### Relevant log output

```shell
NTA
```

### Additional context

_No response_

Abuelodelanada:
## Issue

Closes #208.

`SingletonSnapManager.register()` was only called from `_install_snaps()`, which runs on the `install` and `upgrade-charm` hooks. A lockfile deleted was therefore never re-created.

## Solution

Register at the top of `_reconcile()`, so the lockfiles are treated as desired state.

`register()` prunes only lockfiles whose `unit_name` matches the calling unit, so re-registering does not touch co-located units' entries.

Also drops the autouse `mock_singleton_snap_manager` fixture, which stubbed `get_revisions()` and masked this area from the unit tests. 

### Checklist

## Context

`opentelemetry-collector` and `node-exporter` are machine-wide singleton snaps: one installed
revision per machine, shared by every co-located collector unit. `/opt/singleton_snaps` holds one lockfile per (snap, unit), `LCK..<snap>--rev<N>__<unit>`, which reference-counts units so the last one standing knows it may uninstall the snap on removal.

## Testing Instructions

Unit tests — two added to `tests/unit/test_charm_lifecycle.py`, both verified to fail without the fix:

```
tox run -e unit
```

Manual, following the reproduction steps in #208:

```bash
juju deploy ubuntu
juju deploy opentelemetry-collector otelcol
juju integrate otelcol:juju-info ubuntu:juju-info
```

> With only `juju-info` and no outgoing telemetry relation, `otelcol` is expected to sit in
> `blocked` on missing mandatory relations. That is unrelated to this fix — what matters below is the status *message*, not the status itself.

```bash
# delete a lockfile out of band
juju ssh ubuntu/0 'sudo rm /opt/singleton_snaps/LCK..node-exporter--rev*'

# trigger a hook (the reconcile action is the most reliable trigger)
juju run otelcol/0 reconcile
```

Expected: the lockfile reappears, and the unit does **not** report `Mismatching snap revisions for node-exporter`. Before this change it did, and stayed there through any number of hooks.

To confirm co-tenancy detection is still live, plant a lockfile for a fictitious unit pinning a
newer revision — the unit must then report the mismatch:

```bash
juju ssh ubuntu/0 'sudo touch /opt/singleton_snaps/LCK..node-exporter--rev9999__otelcol_99'
juju run otelcol/0 reconcile
# clean up, or the unit stays blocked
juju ssh ubuntu/0 'sudo rm /opt/singleton_snaps/LCK..node-exporter--rev9999__otelcol_99'
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -188,6 +188,15 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         self._reconcile()
 
     def _reconcile(self):
+        # Re-assert this unit's snap registrations. These lockfiles are reference counts, read
+        # on removal to decide whether the last unit standing may uninstall a shared snap, so
+        # they belong to the desired state. Registering only on install/upgrade-charm left a
+        # lockfile that was deleted out of band missing until the next `juju refresh`.
+        # Refs https://github.com/canonical/opentelemetry-collector-operator/issues/208
+        manager = SingletonSnapManager(self.unit.name)
+        for snap_name in SnapMap.snaps():
+            manager.register(snap_name, SnapMap.get_revision(snap_name))
+
         insecure_skip_verify = cast(bool, self.config.get("tls_insecure_skip_verify"))
         topology = JujuTopology.from_charm(self)
         # NOTE: Only the leader aggregates alerts, to prevent duplication. COS Agent alerts
--- a/src/charm.py
+++ b/src/charm.py
@@ -193,9 +193,7 @@ class OpenTelemetryCollectorCharm(ops.CharmBase):
         # they belong to the desired state. Registering only on install/upgrade-charm left a
         # lockfile that was deleted out of band missing until the next `juju refresh`.
         # Refs https://github.com/canonical/opentelemetry-collector-operator/issues/208
-        manager = SingletonSnapManager(self.unit.name)
-        for snap_name in SnapMap.snaps():
-            manager.register(snap_name, SnapMap.get_revision(snap_name))
+        self._register_snaps()
 
         insecure_skip_verify = cast(bool, self.config.get("tls_insecure_skip_verify"))
         topology = JujuTopology.from_charm(self)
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -188,6 +188,15 @@ def __init__(self, framework: ops.Framework):
...
     def _reconcile(self):
+        # Re-assert this unit's snap registrations. These lockfiles are reference counts, read
+        # on removal to decide whether the last unit standing may uninstall a shared snap, so
+        # they belong to the desired state. Registering only on install/upgrade-charm left a
+        # lockfile that was deleted out of band missing until the next `juju refresh`.
+        # Refs https://github.com/canonical/opentelemetry-collector-operator/issues/208
+        manager = SingletonSnapManager(self.unit.name)
+        for snap_name in SnapMap.snaps():
+            manager.register(snap_name, SnapMap.get_revision(snap_name))
+
```
sed-i: What's the diff b/w this and calliing `self._install_snaps()`?

Abuelodelanada: `_install_snaps()` would also fix the bug, it does the register. The difference is everything it does on every hook.

I extracted the duplicad to to a new private method.

sinapah (approved):

